PR #1, opened 2018-08-20
Add support for mediation layer through setting up servers root as a eureka endpoint

1000TurquoisePogs:
Tested as working - used during SHARE. Room for improvement by supporting websockets and swagger output next.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -9,6 +9,7 @@
   Copyright Contributors to the Zowe Project.
 */
 
+
 'use strict';
 const Promise = require('bluebird');
 const util = require('./util');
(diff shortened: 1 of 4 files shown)


js/plugin-loader.js
```diff
@@ -490,6 +490,15 @@ PluginLoader.prototype = {
   plugins: null,
   pluginMap: null,
   unresolvedImports: null,
+
+  _emitPluginEurekaInfo(pluginDef) {
```
fkovinAtRocket: 1. It doesn't look like "update eureka metedata" is an event at all. What happens here that we need to signal? When the plugin loader has loaded a plugin (either filesystem or dynamic) it already dispatches 'pluginLoaded'. I think even making it *know* that it's invoked by something-Eureka is a violation of modularity
2. This event  isn't even handled anywhere

js/plugin-loader.js
```diff
@@ -644,7 +655,12 @@ PluginLoader.prototype = {
     }
     this.plugins = this._toposortPlugins(pluginContext.plugins, this.pluginMap);
 //    bootstrapLogger.warn('pluginMap empty (plugin-loader.js line530)='
-//        + JSON.stringify(this.pluginMap));    
+//        + JSON.stringify(this.pluginMap));  
+    
+    // Share with index.js the amount of plugins that should load
+    this.emit('givePluginAmount', {
```
fkovinAtRocket: This doesn't look like an event either. In fact, the explicit goal of this design was to make both filesystem and dynamic plugins "first-class citizens" and thus we *can't* know the number of plugins in advance.

If an external caller needs to know whether we're done with loading the filesystem plugins, we can dispatch an event once that's done (and we're now waiting for external plugins)

Also, I know it's just a comment, but even mentioning another module's name here might indicate anti-modular inclinations: we must not care who the listener is, we just implement the contact for the plugin loader. Whether 'index.js' or 'foo.js' or whatever uses us, that shouldn't affect us, we should just do our thing.

js/index.js
```diff
@@ -142,36 +233,68 @@ Server.prototype = {
...
+    
+    // Get number of plugins that should load
+    this.pluginLoader.on('givePluginAmount', event => totalPlugins = event.data);
+    
+    // Add the plugin and add it to the count
+    this.pluginLoader.on('pluginAdded', event => {
+      this.pluginLoaded(event.data);
+      addedPlugins++;            // Make sure total plugins did not get overwritten to null
+
+      if (totalPlugins != null && totalPlugins == addedPlugins) {   // If all the plugins have attempted to load
```
fkovinAtRocket: What happens when a dynamic plugin is added? Do we care if it's registered to Eureka or not?

js/index.js
```diff
@@ -19,8 +20,97 @@ const ProcessManager = require('./process');
 const AuthManager = require('./auth-manager');
 const WebAuth = require('./webauth');
 const unp = require('./unp-constants');
+const eureka = require('eureka-js-client').Eureka;
+const http = require('http');
+
 
 
+
+const MEDIATION_LAYER_INSTANCE_DEFAULTS = {
```
fkovinAtRocket: Well, the idea behind index.js was that it wires various pieces together but doesn't contain much logic itself. Mediation layer support is worth a separate module IMO.

1000TurquoisePogs: Yes that would be fine, but it still needs to act upon the plugin loading events.

js/index.js
1000TurquoisePogs: I don't think the mediation layer currently supports this, but it would be good.

js/plugin-loader.js
1000TurquoisePogs: You must know the number of initial plugins for certain things (import resolution, eureka handshake) so an event was created. Dynamic adding should be delayed until after this point to avoid a problem in counting but it happens very quickly currently so it hasnt yet posed a problem.

js/plugin-loader.js
1000TurquoisePogs: Yeah this may be dead code. There was more that was removed  from the original PR due to not being useful to the mediation layer than what actually was merged

PR #8, opened 2018-09-07
Error reporting API in the node server

fkovinAtRocket:
Add a utility method that makes error objects. Make the config service
use it.

Signed-off-by: Fyodor Kovin <fkovin@rocketsoftware.com>


js/util.js
```diff
@@ -109,6 +109,27 @@ module.exports.readFilesToArray = function(fileList) {
   }
 };
 
+const errorProto = {
+    "_objectType": "com.zoeproject.error",
```
1000TurquoisePogs: `org.zowe.zlux.error` is the best for now.

js/util.js
```diff
@@ -109,6 +109,27 @@ module.exports.readFilesToArray = function(fileList) {
...
+    "_metaDataVersion": "1.0.0",
+    "returnCode": "1",
+    "messageID": "ZOE000E",
+    "messageTemplate": "An error occurred",
+    "messageParameters": {},
+    "messageDetails": "An error occurred when processing the request"
+  };
+
+module.exports.makeErrorObject = function makeError(details) {
```
1000TurquoisePogs: Please provide this function via the context object of a dataservice: https://github.com/zowe/zlux-proxy-server/blob/master/plugins/config/lib/configService.js#L2567

This way, dataservice writers would not need to use a (relative pathed!) require() in order to use it: it would just be there already.

fkovinAtRocket:
Replaced by https://github.com/zowe/zlux-proxy-server/pull/46

PR #9, opened 2018-09-17
Import resolution failure bugfix

1000TurquoisePogs:
Import resolution failure would not wait for dependencies to resolve, causing import failure. This fixes it

Signed-off-by: Sean Grady <sgrady@rocketsoftware.com>

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -240,7 +240,7 @@ Server.prototype = {
     // Add to eureka metadata info
 
     var totalPlugins = -1;    // Stores the number of plugins that should load
-    var addedPlugins = 0;     // Stores the number of plugins that have attempted to load
+    var visitedPlugins = 0;     // Stores the number of plugins that have attempted to load
     
     // Get number of plugins that should load
     this.pluginLoader.on('givePluginAmount', event => totalPlugins = event.data);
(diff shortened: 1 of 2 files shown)


js/index.js
```diff
@@ -246,30 +244,29 @@ Server.prototype = {
...
+      pluginsLoaded.push(event.data);
+      this.pluginLoaded(event.data).then(() => {
+        addedPlugins++;            // Make sure total plugins did not get overwritten to null
+        installLogger.info('Installed plugin: ' + event.data.identifier);
+        if (totalPlugins != null && totalPlugins == addedPlugins) {   // If all the plugins have attempted to load
+          this.finishPluginInitialization(pluginsLoaded, webAppOptions);
         }
-      }
+      }, err => {
+        addedPlugins++;
```
rpenny125: "addedPlugins" sounds misleading if it counts plugins that weren't added. Perhaps visitedPlugins?

js/webapp.js
```diff
@@ -783,6 +783,37 @@ WebApp.prototype = {
...
+        this._resolveImports(plugin, urlBase);
+      } catch (e) {
+        unresolvedPlugins.push(plugin);
+        //keep in array and retry
+      }
+    });
+    if (unresolvedPlugins.length === 0) {
+      installLog.info(`All imports resolved for all plugins.`);
+      return true;
+    } else if (unresolvedPlugins.length !== pluginDefs.length) {
```
rpenny125: It's not immediately obvious to me why you should need to re-drive this method now that you've pushed this to the post-processing phase (i.e. after all plugins have been initially loaded). So, if it really is needed, I think a comment would be helpful.

How well is an import loop handled? Does this algorithm terminate? I think it does, but did you test? Maybe later we can worry about being able to report errors of that type in more detail. For now it's enough that this algorithm terminates, and the full list of failures is apparent in the log.

1000TurquoisePogs: let me retest this to see: i was concerned about imports-of-imports-of-imports even if it is not a recommended thing to do. I'll try to make the react app import the iframe app dataservice which imports the angular dataservice and see what the behavior is in this function.

rpenny125:
NOTE: I am merging, even though testing exposed (existing) bug https://github.com/zowe/zlux-proxy-server/issues/10

1000TurquoisePogs:
"So how did you create the situation when you needed to retry something?"
Imports were not working for any of the following:
1. Trying to import [tn3270data](https://github.com/zowe/zlux-proxy-server/blob/master/plugins/terminal-proxy/pluginDefinition.json#L27) into [tn3270-ng2](https://github.com/zowe/tn3270-ng2/blob/master/pluginDefinition.json)
2. Trying to import [hello](https://github.com/zowe/sample-angular-app/blob/master/pluginDefinition.json#L26) from sample-angular-app into [sample-iframe-app](https://github.com/zowe/sample-iframe-app/blob/master/pluginDefinition.json)
3. Trying to import an import... [hello from sample-iframe-app](https://github.com/zowe/sample-iframe-app/blob/master/pluginDefinition.json) into [sample-react-app](https://github.com/zowe/sample-react-app/blob/master/pluginDefinition.json)

The last one isn't really important, but I found that it was possible to do after #9 
I know that around the time this was made there was an urgent deadline, so probably Robert was involved because he was in the office and was having a similar problem. 

I think the reason the imports were not working was due to asynchronous router loading... the routers weren't available at the time the import resolution logic was being run.

When plugins are being dynamically loaded over REST, it's easy to say that it's impossible to satisfy an import because you have just asked for 1 plugin, and maybe you have a way of knowing that there's none currently being processed in the pipeline, and so that if you can't find the import, then it's just not something that can be satisfied. 
But, for plugins that are loaded at startup, you have a full list to process initially and are able to say that the imports should be satisfy-able, as long as the code is structured to wait for the router to be available before trying to import it.

PR #13, opened 2018-09-28
Added additional redirection logic

jordanfilteau1995:
Proxy Server will now:
- try to redirect 404's using the "Referer" header.
- proxy better

 

Signed-off-by: Jordan Filteau <jfilteau@rocketsoftware.com>

### DIFF
--- a/js/webapp.js
+++ b/js/webapp.js
@@ -462,13 +462,8 @@ WebApp.prototype = {
   },
   
   makeExternalProxy(host, port, urlPrefix, isHttps, noAuth, pluginID, serviceName) {
-    installLog.info("----------------------------------");
-    installLog.info("PluginID: " + pluginID);
-    installLog.info("ServiceName: " + serviceName);
-    installLog.info("----------------------------------");
     const r = express.Router();
     installLog.info(`Setting up proxy to ${host}:${port}/${urlPrefix}`);
-    installLog.info('----------------------PROXY---------------------');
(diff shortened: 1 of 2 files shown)


jordanfilteau1995:
@1000TurquoisePogs

js/webapp.js
```diff
@@ -884,8 +889,37 @@ WebApp.prototype = {
...
+         */
+        if (header == 'referer') {
+          var pattern = /^http.+\/ZLUX\/plugins\/.+/;
+          if (pattern.test(headers[header])) {
+            var parts = headers[header].split("/");
+            var zluxIndex = parts.indexOf("ZLUX");
+            var pluginID = parts[zluxIndex + 2];
+            var serviceName = parts[zluxIndex + 4];
+            var myProxy = proxyMap.get(pluginID + ":" + serviceName);
+            var fullUrl = req.originalUrl;
```
1000TurquoisePogs: This part of the code has one area where it falls short. Basically, if I went to http://localhost:12360/ZLUX/plugins/com.rs.zosmf.proxy/services/data instead of http://localhost:12360/ZLUX/plugins/com.rs.zosmf.proxy/services/data**/** then the path lookup for HTML would omit the **data** part of the URL. 

It would be good to try and detect relative pathing to a minor degree to see that if the requested URL appears to be some subject of the referrer URL, then likely this is a relative path gone awry.

This is not really a bug for the PR in that this behavior existed regardless, so I'll still merge.

image

PR #14, opened 2018-10-05
MVD-206: service versions, unit tests

fkovinAtRocket:
Unit testing: just run `mocha <test>.js`, you'll need to run `npm install chai chai-http`  to run the webapp tests.


js/depgraph.js
```diff
@@ -0,0 +1,279 @@
...
+      }
+      pluginNode.visiting = false;
+      pluginNode.visited = true;
+      pluginNode.finishingTime = ++time;
+      logger.debug(`${pluginNode.pluginId}: `
+          + `${pluginNode.discoveryTime}/${pluginNode.finishingTime}`);
+      //See the proof at the end of Cormen et al. (2001), 
+      // "Section 22.4: Topological sort"
+      //loop invariant derived from the proof
+      assert((pluginsSorted.length === 0)
```
fkovinAtRocket: Perhaps it will be better to comment it out in production. the thing with assert in JS is that an AssertionError is just an error, that will be caught by *any* catch block. So, most likely, it will just confuse the program in some place and won't terminate the process anyway.

1000TurquoisePogs:
resolves zowe/zlux#8

js/plugin-loader.js
```diff
@@ -438,15 +445,15 @@ ProxyConnectorPlugIn.prototype = {
...
-      bootstrapLogger.warn("Authentication plugin was found which was not requested in "
-          + "the server configuration file's dataserviceAuthentication object. "
-          + "Skipping load of this plugin");
-      return false;
-    }
+//    //we should not load authentication types that are 
+//    //not requested by the administrator
+//    if (!context.authManager.authPluginRequested(this.identifier,
+//      this.authenticationCategory)) {
+//      bootstrapLogger.warn("Authentication plugin was found which was not requested in "
```
1000TurquoisePogs: Why was this commented out? we were actually using this.

js/plugin-loader.js
```diff
@@ -43,113 +44,75 @@ const defaultOptions = {
...
-    let success = true;
-    for (const plugin of plugins) {
-      const externalImports = this.map[plugin.identifier];
-      if (externalImports) {
-        if (plugin.dataServices) {
-          for (const dataservice of plugin.dataServices) {
-            dataservice.externalRefs = externalImports[dataservice.name];
-            delete externalImports[dataservice.name];
-          }
+    if (this.versionRequirements) {
```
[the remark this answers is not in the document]
fkovinAtRocket: See your own comment here, notice the "requiredServices" property (I slightly reorganized/rephrased it but the idea is the same)
https://jira.rocketsoftware.com/browse/MVD-206?focusedCommentId=7152355&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-7152355

js/package.json
```diff
@@ -16,12 +16,12 @@
   "dependencies": {
     "bluebird": "~3.5.1",
     "body-parser": "~1.18.3",
+    "eureka-js-client": "~4.4.1",
     "express": "~4.16.3",
     "express-session": "~1.15.6",
     "express-ws": "~4.0.0",
-    "eureka-js-client": "~4.4.1",
     "require-from-string": "~2.0.2",
-    "toposort": "~2.0.2",
```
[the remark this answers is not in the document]
fkovinAtRocket: The main thing here is recursively removing the dependents of a broken plugin. I first wanted to do it in one pass, together with toposorting the nodes. Though I later found two passes more reasonable, the algorithms and implementations are very similar, so I decided, importing toposort won't give us much. 

The toposort algorithm is just from Cormen (as mentioned below) and also quoted on Wikipedia.

js/plugin-loader.js
fkovinAtRocket: It looked misplaced: why does this have to be under the "proxy connector"?

js/plugin-loader.js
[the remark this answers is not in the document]
1000TurquoisePogs: You must have confused me back then because I'm confused now. Version requirement marking is important for imports, but for other dataservice types it seems that while you can make this relationship statement it wouldn't add value. A plugin is developed to work as a coherent unit, so doing version checks from one service to another in the same plugin seems pointless in terms of the compatibility already being guaranteed by the developer as a result of publishing the plugin.

Let's say, for example, that service foo had its API updated, such that the new service was able to serve APIs v1 and v2. And let's say that service bar in the same plugin relies on calling foo @ v1. Does this relationship really need to be stated? Isn't it implied that a given release of a plugin is valid by the developer releasing it as a whole, rather than in parts? I mean, if bar needed v1 but foo only provided v2, it would not work but specifying that additional metadata wouldn't solve the problem. People would just issue a bug report which would lead to the fix instead.

However - you do of course need to do version requirement checks for IMPORT dataservices, since you should only be able to import when the version pattern matches, since a given plugin expects a service of a certain version, which is obtained externally and therefore has a higher chance of failure due to out of date plugins.

So, for an import, imagine:
    {
      "version": "//makes sense to be inherited by the import resolution",
      "type": "import",
      "sourceName": "hello",
      "localName": "hello",
      "sourcePlugin": "org.zowe.zlux.sample.iframe",
      "versionRequirements": "~3.0.0"
    },

js/plugin-loader.js
1000TurquoisePogs: Nevermind, you are right. Its a duplicate block and I was thinking of the duplicate: https://github.com/zowe/zlux-proxy-server/blob/e66c15d44ae5632dfb7fd8a37849e3aea41856eb/js/plugin-loader.js#L404

js/plugin-loader.js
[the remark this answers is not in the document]
fkovinAtRocket: Considering this scenario: there's service A in two versions: v1 and v2, and there's service B who wants to call A, not the most recent version, but v1 instead.  I don't find this scenario impossible. I don't see why this is necessarily a bug: what if they deliberately don't want to update B to the newer version of A's API for some reason? Furthermore, if in the situation of B calling A we just choose the most recent  version of A, that would simply be an arbitrary choice on our part: why do we think that newest is the best? 

That was exactly the question back then: should we actually bother to implement those "local version requirements" or not? I thought it's worth it, but if you think that's useless, probably maintaining that code would be a burden and I'd need to remove it.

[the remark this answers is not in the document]
1000TurquoisePogs: OK, you raise a good point about versioning for non-import services but I'm having trouble finding the code that could do this: if B depends on a version of A, callService should either have a version parameter or know that a named service A requested by B corresponds to a particular version of A due to the stated versionRequirements. There's an injectible middleware piece for associating proxies with callService but it doesn't look like it has changed to accommodate versioning.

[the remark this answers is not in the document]
fkovinAtRocket: I don't get your question: one part is done in the code we're discussing right here, the other part is indeed in callService(): https://github.com/zowe/zlux-proxy-server/pull/14/files#diff-55d2935735e0c392902cca19a95e92c0R299

fkovinAtRocket:
It looks like I've resolved the conflicts.

Also, I've merged with the  APINL branch here just in case: https://github.com/fkovinAtRocket/zlux-proxy-server-1/pull/1

fkovinAtRocket:
I'm getting these errors now:
```
[2018-12-06 10:06:27.806 _unp.bootstrap WARNING] - Could not initialize plugin com.zowe.explorer.server.auth: REQUIRED_SERVICE_VERSION_NOT_FOUND
[2018-12-06 10:06:27.806 _unp.bootstrap WARNING] - Could not initialize plugin org.zowe.terminal.tn3270: REQUIRED_SERVICE_VERSION_NOT_FOUND
[2018-12-06 10:06:27.806 _unp.bootstrap WARNING] - Could not initialize plugin org.zowe.terminal.vt: REQUIRED_SERVICE_VERSION_NOT_FOUND
[2018-12-06 10:06:27.806 _unp.bootstrap WARNING] - Could not initialize plugin org.zowe.zlux.auth.zosmf: REQUIRED_SERVICE_VERSION_NOT_FOUND
[2018-12-06 10:06:27.806 _unp.bootstrap WARNING] - Could not initialize plugin org.zowe.zlux.sample.react: REQUIRED_SERVICE_NOT_FOUND
[2018-12-06 10:06:27.806 _unp.bootstrap WARNING] - Could not initialize plugin org.zowe.zlux.sample.iframe: REQUIRED_SERVICE_VERSION_NOT_FOUND
[2018-12-06 10:06:27.809 _unp.bootstrap WARNING] - Service org.zowe.zosmf.workflows::zosmf doesn't specify a version. Resorting to the plugin's API version 1.0.0
[2018-12-06 10:06:27.813 _unp.bootstrap WARNING] - Service org.zowe.zosmf.workflows::zosmftracker doesn't specify a version. Resorting to the plugin's API version 1.0.0
[2018-12-06 10:06:27.816 _unp.bootstrap WARNING] - Service org.zowe.zlux.sample.angular::hello doesn't specify a version. Resorting to the plugin's API version 1.0.0
[2018-12-06 10:06:27.822 _unp.bootstrap WARNING] - Service org.zowe.zlux.ng2desktop::browser-preferences doesn't specify a version. Resorting to the plugin's API version 1.0.0
[2018-12-06 10:06:27.908 _unp.bootstrap WARNING] - Service org.zowe.zlux.proxy.zosmf::data doesn't specify a version. Resorting to the plugin's API version 1.0.0
[2018-12-06 10:06:27.912 _unp.bootstrap WARNING] - Service org.zowe.zossystem.subsystems::data doesn't specify a version. Resorting to the plugin's API version 1.0.0
[2018-12-06 10:06:27.913 _unp.bootstrap WARNING] - Service org.zowe.terminal.proxy::tn3270data doesn't specify a version. Resorting to the plugin's API version 1.0.0
[2018-12-06 10:06:27.926 _unp.bootstrap WARNING] - Service org.zowe.terminal.proxy::tn5250data doesn't specify a version. Resorting to the plugin's API version 1.0.0
[2018-12-06 10:06:27.927 _unp.bootstrap WARNING] - Service org.zowe.terminal.proxy::vtdata doesn't specify a version. Resorting to the plugin's API version 1.0.0
[2018-12-06 10:06:27.931 _unp.bootstrap WARNING] - Service org.zowe.configjs::data doesn't specify a version. Resorting to the plugin's API version 1.0.0
[2018-12-06 10:06:27.933 _unp.bootstrap WARNING] - Service com.zowe.explorer.server.proxy::data doesn't specify a version. Resorting to the plugin's API version 1.0
```
Looks like some plugin defs need to be fixed.

fkovinAtRocket:
https://github.com/zowe/explorer-server-auth/pull/5
https://github.com/zowe/sample-angular-app/pull/10
https://github.com/zowe/sample-react-app/pull/7
https://github.com/zowe/sample-iframe-app/pull/8
https://github.com/zowe/tn3270-ng2/pull/11
https://github.com/zowe/vt-ng2/pull/9
https://github.com/zowe/zlux-workflow/pull/20
https://github.com/zowe/zos-subsystems/pull/12
https://github.com/zowe/zosmf-auth/pull/7

fkovinAtRocket:
https://github.com/fkovinAtRocket/db-browser-db2/pull/1

1000TurquoisePogs:
Testing was successful, but I dnd't get to test everything, so I will check with others tomorrow & run your tests before merging just to be sure.

test/webapp/test-plugin/pluginDefinition.json
```diff
@@ -0,0 +1,37 @@
+{
+  "identifier": "com.rs.testplugin",
+  "baseURI": "test",
```
1000TurquoisePogs: baseURI is vestigial - please remove it to prevent confusion.

test/webapp/test-plugin/pluginDefinition.json
```diff
@@ -0,0 +1,37 @@
+{
+  "identifier": "com.rs.testplugin",
```
1000TurquoisePogs: org.zowe please, not com.rs

test/plugin-loader/plugin-loader.js
```diff
@@ -0,0 +1,247 @@
...
+          {
+            "type": "router",
+            "name": "foo",
+            "fileName": "nop-router.js",
+            "version": "1.2.3"
+          },
+          {
+            "type": "router",
+            "name": "foo",
+            "fileName": "nop-router.js",
```
1000TurquoisePogs: I don't think you can actually access foo@1.2.3 and foo@4.5.6 in this situation. If you have 2 routers which come from the same file, wouldn't you need to specify individual & different factory function names with `routerFactory`? Otherwise the returned object is identical.

fkovinAtRocket: Yes, but that service is never invoked, so that's irrelevant for this test's purposes

PR #15, opened 2018-10-23
Internationalization of plugin definitions

lchudinov:
Test with: 
* https://github.com/zowe/sample-angular-app/pull/7
* https://github.com/zowe/zos-subsystems/pull/5

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -22,7 +22,7 @@
     "express": "~4.16.3",
     "express-session": "~1.15.6",
     "express-ws": "~4.0.0",
-    "glob": "^7.1.3",
+    "glob": "~7.1.3",
     "require-from-string": "~2.0.2",
     "toposort": "~2.0.2",
     "ws": "~5.2.0"
--- a/js/package.json
+++ b/js/package.json
@@ -14,12 +14,15 @@
     }
   ],
   "dependencies": {
+    "accept-language-parser": "^1.5.0",
     "bluebird": "~3.5.1",
     "body-parser": "~1.18.3",
+    "cookie-parser": "^1.4.3",
+    "eureka-js-client": "~4.4.1",
     "express": "~4.16.3",
     "express-session": "~1.15.6",
     "express-ws": "~4.0.0",
-    "eureka-js-client": "~4.4.1",
+    "glob": "^7.1.3",
     "require-from-string": "~2.0.2",
     "toposort": "~2.0.2",
     "ws": "~5.2.0"
--- a/js/package.json
+++ b/js/package.json
@@ -14,7 +14,7 @@
     }
   ],
   "dependencies": {
-    "accept-language-parser": "^1.5.0",
+    "accept-language-parser": "~1.5.0",
     "bluebird": "~3.5.1",
     "body-parser": "~1.18.3",
     "cookie-parser": "^1.4.3",
--- a/js/translation-utils.js
+++ b/js/translation-utils.js
@@ -18,7 +18,7 @@ const acceptLanguageParser = require('accept-language-parser');
 const utilLog = zluxUtil.loggers.utilLogger;
 
 /* loadTranslations reads pluginDefinition.i18n.{lang}-{Country}.json files
-/  for a giveen plugin location.
+/  for a given plugin location.
 / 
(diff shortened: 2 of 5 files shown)


js/package.json
```diff
@@ -14,12 +14,15 @@
...
+    "accept-language-parser": "^1.5.0",
     "bluebird": "~3.5.1",
     "body-parser": "~1.18.3",
+    "cookie-parser": "^1.4.3",
+    "eureka-js-client": "~4.4.1",
     "express": "~4.16.3",
     "express-session": "~1.15.6",
     "express-ws": "~4.0.0",
-    "eureka-js-client": "~4.4.1",
+    "glob": "^7.1.3",
```
1000TurquoisePogs: Please make these ~ versions instead of ^ versions, for ease of maintenance over time.

lchudinov: Done

js/translation-utils.js
```diff
@@ -0,0 +1,152 @@
...
+
+// getAcceptLanguageFromCookies builds acceptLanguage string
+// based on user preferences stored in cookies
+function getAcceptLanguageFromCookies(cookies) {
+  const prefix = 'org.zowe.zlux.zlux-app-manager.preferences';
+  const languageKey = `${prefix}.language`;
+  const localeKey = `${prefix}.locale`;
+  const language = cookies[languageKey];
+  const locale = cookies[localeKey];
+  if (language && locale) {
```
rpenny125: "language" *could* be es-MX and locale *could* be US (for example) for a spanish speaker raised in Mexico, but currently living or visiting in the US.

Should either just use language, or check if it already has both language and locale before adding the locale.

js/package.json
```diff
@@ -14,12 +14,15 @@
     }
   ],
   "dependencies": {
+    "accept-language-parser": "^1.5.0",
```
rpenny125: I think that Sean's comment about ^ vs ~ applies here, too.

js/translation-utils.js
```diff
@@ -0,0 +1,152 @@
...
+  const webContent = pluginDefClone.webContent;
+  const translationMap = translationMaps[langCountry];
+  translateObject(webContent, translationMap);
+  return pluginDefClone;
+}
+
+// getAcceptLanguageFromCookies builds acceptLanguage string
+// based on user preferences stored in cookies
+function getAcceptLanguageFromCookies(cookies) {
+  const prefix = 'org.zowe.zlux.zlux-app-manager.preferences';
```
rpenny125: Seeing this string (used in zlux-app-manager) here also, makes me wonder if the "cookie-setting service" that is currently in virtual-desktop (https://github.com/zowe/zlux-app-manager/blob/master/virtual-desktop/nodeServer/src/browserPreferences.ts) should be moved into zlux-proxy service (would involve a client change also, since the URI would have to change).

Sean, what do you think?

1000TurquoisePogs: Yes it is wrong, but that should be done as a separate work item. I would like to see the whole lang feature built right into the server (as in, not even a plugin) but would accept it being a plugin within zlux-proxy-server/plugins short term.

1000TurquoisePogs:
I will test later today and hopefully merge.

js/translation-utils.js
```diff
@@ -0,0 +1,157 @@
...
+  const translationMaps = {};
+  const relativePath = 'web/assets/i18n';
+  const filePrefix = 'pluginDefinition.i18n.';
+  const fileExt = '.json';
+  const pattern = path.join(
+    pluginLocation,
+    relativePath,
+    `${filePrefix}*${fileExt}`
+  );
+  const files = glob.sync(pattern, {});
```
1000TurquoisePogs: For the record: yes this is a synchronous call but no it shouldn't really be bad since this function is done one per plugin at startup. I feel that we have not captured all scenarios in which we will load plugins after startup, but it seems like we'd want to avoid loading plugins at scale, such that this again shouldn't impact performance.

1000TurquoisePogs:
I tested that the language you choose is respected by this code, and it results in the /plugins API sending translated responses such that the UI shows plugin names as translated.

PR #18, opened 2018-10-24
Bug/MVD-947v2 Plugin Defintion Validation

timgerstel:
Validates plugin definitions using the AJV package which allows for the creation of JSON schemas.

Replaces this pull request: [#33](https://github.com/gizafoundation/zlux-proxy-server/pull/33)

The following must be merged with this pull request:
gizafoundation/zosmf-auth#3
gizafoundation/workshop-user-browser-app#5

Potential TODO: Have warnings logged instead of errors thrown for certain validation failures (i.e. semantic versioning)

(note from rpenny)
According to the above text, this replaces pull #24 . That PR indicated committing these other PR's so I am listing them here before I close #24.:
DO NOT MERGE THIS FILE WITHOUT ALSO MERGING:
gizafoundation/zosmf-auth#3
gizafoundation/sample-app#9
gizafoundation/tn3270-ng2#10
gizafoundation/zlux-app-manager#35
gizafoundation/zos-subsystems#3
gizafoundation/zss-auth#3
gizafoundation/zlux-shared#6
gizafoundation/sample-iframe-app#3
gizafoundation/workshop-starter-app#2
gizafoundation/workshop-user-browser-app#5

EVERYTHING MUST BE MERGED AT THE SAME TIME,


js/Schemas/pluginDefinitionSchema.json
```diff
@@ -0,0 +1,345 @@
...
+         },
+         "then": {
+            "properties": {
+               "webContent": {
+                  "if": {
+                     "additionalProperties": false
+                  },
+                  "then": {},
+                  "else": {
+                     "webContentSchema",
```
1000TurquoisePogs: can you describe what webContentSchema is doing here? I can't. It's definitely invalid JSON so I'm thinking this should just be removed.

js/Schemas/pluginDefinitionSchema.json
```diff
@@ -0,0 +1,345 @@
...
+         "properties": {
+            "pluginType": {
+               "const": "nodeAuthentication"
+            }
+         }
+      },
+      "then": {
+         "required": [
+            "authenticationCategory",
+            "filename"
```
1000TurquoisePogs: "filename" is fine except that it breaks everything because we incorrectly were using "fileName". Can you fix configjs and terminalproxy within this same pull request? We'll just need to open up other pull requests to fix the other repos dependent upon this one.

js/package.json
```diff
@@ -14,6 +14,8 @@
     }
   ],
   "dependencies": {
+    "ajv": "^6.5.2",
```
1000TurquoisePogs: Everything that goes into package.json must use the "~" sign instead of the "^" sign because it locks things down to patch changes rather than more major (possibly breaking) changes.

lib/schema/pluginDefinitionSchema2.json
```diff
@@ -0,0 +1,289 @@
+{
+    $schema: "http://json-schema.org/schema#",
```
1000TurquoisePogs: Everything in this file is missing double-quotes around the attributes... this can't be valid JSON?

lib/schema/webContentSchema2.json
```diff
@@ -0,0 +1,98 @@
+{
+    $schema: "http://json-schema.org/schema#",
```
1000TurquoisePogs: Again this can't be a valid JSON file; the attributes are missing double-quotes.

js/Schemas/pluginDefinitionSchema.json
1000TurquoisePogs: I am guessing the intent was to refer to one schema from another. I think this page describes the correct way to do it: https://json-schema.org/understanding-json-schema/structuring.html#extending

lib/schema/Schemas.js
```diff
@@ -0,0 +1,396 @@
...
+            "errorMessage": {
+                "type": "must be a string"
+            }
+        },
+        "pluginVersion": {
+            "type": "string",
+            "pattern": semverRegex,
+            "errorMessage": {
+                "type": "must be a string",
+                "pattern": "must follow semantic versioning spec (x.y.z-anything)"
```
1000TurquoisePogs: Please cleanup the logs here, I'm seeing a bit of redundant info. I intentionally used the wrong version to cause a schema problem, and see this:

```
[2018-12-04 14:32:07.768 _unp.bootstrap INFO] - Validating Plugin Definition at ../../explorer-server-auth//pluginDefinition.json
[2018-12-04 14:32:07.790 _unp.bootstrap INFO] - Loaded Plugin Defintion Schema
Error: Key "pluginVersion" must follow semantic versioning spec (x.y.z-anything), is 1.0
    at Object.validatePluginDef (C:\Users\sgrady\Downloads\zlux-404\zlux-proxy-server\js\pluginDefinitionValidation.js:73:15)
    at PluginLoader._readPluginDef (C:\Users\sgrady\Downloads\zlux-404\zlux-proxy-server\js\plugin-loader.js:548:32)
    at PluginLoader.loadPlugins (C:\Users\sgrady\Downloads\zlux-404\zlux-proxy-server\js\plugin-loader.js:598:29)
    at Server.<anonymous> (C:\Users\sgrady\Downloads\zlux-404\zlux-proxy-server\js\index.js:270:23)
    at Generator.next (<anonymous>)
    at Generator.tryCatcher (C:\Users\sgrady\Downloads\zlux-404\zlux-proxy-server\js\node_modules\bluebird\js\release\util.js:16:23)
    at PromiseSpawn._promiseFulfilled (C:\Users\sgrady\Downloads\zlux-404\zlux-proxy-server\js\node_modules\bluebird\js\release\generators.js:97:49)
    at Server.start (C:\Users\sgrady\Downloads\zlux-404\zlux-proxy-server\js\node_modules\bluebird\js\release\generators.js:201:15)
    at Object.<anonymous> (C:\Users\sgrady\Downloads\zlux-404\zlux-example-server\js\zluxServer.js:18:13)
    at Module._compile (internal/modules/cjs/loader.js:688:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:699:10)
    at Module.load (internal/modules/cjs/loader.js:598:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:537:12)
    at Function.Module._load (internal/modules/cjs/loader.js:529:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:741:12)
    at startup (internal/bootstrap/node.js:285:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:739:3)
[2018-12-04 14:32:07.801 _unp.bootstrap WARNING] - Error: Key "pluginVersion" must follow semantic versioning spec (x.y.z-anything), is 1.0
[2018-12-04 14:32:07.801 _unp.bootstrap INFO] - Failed to load com.zowe.explorer.server.auth.json
```

1. Don't need the same message twice
2. Don't need a stack trace at all (You can solve 1 and 2 here but eliminating the non-_unp.bootstrap log and then this PR gets rid of that stack trace: https://github.com/zowe/zlux-proxy-server/pull/33)
3. Do need to know which plugin (by ID) failed on the same line that prints the reason for failure. I know that the info is a few lines above and below, but when people send you error messages, they do not always expect that, and therefore having the info in the same message can reduce common questions & mistakes.

lib/schema/Schemas.js
```diff
@@ -0,0 +1,396 @@
+const semverRegex = "^[0-9]+\.[0-9]+\.[0-9]+(-.+)?$"
+
+const pluginDefinitionSchema = Object.freeze({
```
1000TurquoisePogs: This has no notion of version. We're going to make improvements over time, new features, and so on. So, the pluginDefinition schema will change by version. Each plugin contains an attribute, `apiVersion`, which describes which schema it is attempting to conform to. So, what you have here is the schema for 1.0.0.
You need the server to be able to take multiple schemas, and test each plugin against the schema definition that matches the one it states it is using.

js/pluginDefinitionValidation.js
```diff
@@ -0,0 +1,127 @@
...
+
+const Schemas = require('../lib/schema/Schemas');
+const pluginDefinitionSchema = Schemas.pluginDefinitionSchema;
+
+// **************************************************************************************************************************************************
+// Validates the plugin definition based on the json schema defined above
+// Uses custom error messages in conjunction with the error messages defined in the schema
+
+module.exports.validatePluginDef = function validatePluginDef(pluginDef) {
+    const validate = ajv.compile(pluginDefinitionSchema)
```
1000TurquoisePogs: You should not be compiling per plugin def. Compile each schema that exists once.
See example from https://www.npmjs.com/package/ajv
```
var validate = ajv.compile(schema);
console.log(validate(2.01)); // true
console.log(validate(3.99)); // true
console.log(validate(2)); // false
console.log(validate(4)); // false
```

lib/schema/Schemas.js
```diff
@@ -0,0 +1,396 @@
...
+            "then": {
+                "properties": {
+                    "webContent": {
+                        "if": {
+                            "additionalProperties": false //Checks if the webcontent is empty
+                        },
+                        "then": {},
+                        "else": {
+                            "allOf": [
+                                { "$ref": "#/definitions/webContent" },
```
1000TurquoisePogs: A previous requirement of this work was that the definitions themselves be JSON, and here's why: if we run into a situation where someone needs a plugin that uses a schema version that the server doesnt have, you can provide a JSON file harmlessly. But, sending someone a JS file can be a security concern for many. 

Either way, what must be done is that each schema must be in its own JSON/JS file, such that we can flexibly validate revisions to the spec for the plugins that will use it.

fkovinAtRocket:
I'd take a look at this: https://github.com/YousefED/typescript-json-schema

I didn't spend much time researching this, just quickly searched and found this. But if it's what it appears to be and it works, we might kill two birds with one stone:
- having type script defs for data is good because, obviously, you can write in type script
- this whole JSON schema appears to be very XML-ish in that it's very verbose and tedious, looks quite error-prone. Writing type script defs instead would be a lot more pleasant and efficient.

fkovinAtRocket:
Also, plugin defs are not the only thing to validate: we also have configuration, which is actually a lot more likely to get edited by people outside the team. I'm getting reports that errors in the config can be very puzzling and time confusing, like, you say "mvdPort" instead of "zssPort"

PR #28, opened 2018-11-15
Feature/add sync method to get key

janiago:
In my local project which includes Zowe, I need a function like getKeyFromPassword(), but I need it return a result instead of running a callback. 
So I added a new function which works in sync way

### DIFF
--- a/js/plugin-loader.js
+++ b/js/plugin-loader.js
@@ -326,6 +326,10 @@ Plugin.prototype = {
             else {
               throw new Error (`No file name for data service`)
             }
+            // Make the relative path clear. process.cwd() is zlux-example-server/bin/
+            if (!path.isAbsolute(fileLocation)) {
+              fileLocation = path.join(process.cwd(),fileLocation);
+            }
             const nodeModule = require(fileLocation);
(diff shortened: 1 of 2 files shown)


js/plugin-loader.js
```diff
@@ -326,6 +326,10 @@ Plugin.prototype = {
             else {
               throw new Error (`No file name for data service`)
             }
+            // Make the relative path clear. process.cwd() is zlux-example-server/bin/
+            if (!path.isAbsolute(fileLocation)) {
```
fkovinAtRocket: Can you please provide some details as to why this is needed? What exactly did you do so that this makes a difference?

I'm asking because I've made a similar change since this PR was created

js/plugin-loader.js
```diff
@@ -326,6 +326,10 @@ Plugin.prototype = {
...
+              fileLocation = path.join(process.cwd(),fileLocation);
```
fkovinAtRocket: Can this even be right? Will a data service ever be found under just process.cwd(), which is, as you're saying, 'zlux-example-server/bin/' ?

PR #29, opened 2018-11-15
Feature/pfx passphrase

JerryGaoRocket:
The function to ask for password of PFX key didn't work. 
I've fixed the issue.

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -23,6 +23,7 @@
     "express-session": "~1.15.6",
     "express-ws": "~4.0.0",
     "glob": "~7.1.3",
+	"readline-sync": "^1.4.9",
     "require-from-string": "~2.0.2",
     "toposort": "~2.0.2",
     "ws": "~5.2.0"
--- a/js/webserver.js
+++ b/js/webserver.js
@@ -101,7 +101,7 @@ WebServer.prototype = {
     }
   },
 
-  startListening: Promise.coroutine(function* (app) {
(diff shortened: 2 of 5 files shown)


js/package.json
```diff
@@ -23,6 +23,7 @@
     "express-session": "~1.15.6",
     "express-ws": "~4.0.0",
     "glob": "~7.1.3",
+	"readline-sync": "^1.4.9",
```
1000TurquoisePogs: This needs to start with "~" instead of "^", since we want our versions to vary by no more than a patch version to maintain consistency over time.

js/webserver.js
```diff
@@ -112,14 +112,10 @@ WebServer.prototype = {
           this.expressWsHttps = expressWs(app, this.httpsServer, {maxPayload: 50000});
           listening = true;
         } catch (e) {
-          if (e.message == 'mac verify failure' && !noPrompt) {
-            const r = reader();
-            try {
-              httpsOptions.passphrase = yield reader.readPassword(
+          if (e.message == 'mac verify failure') {
+            const r = new reader();
+            this.httpsOptions.passphrase = r.readPasswordSync(
```
1000TurquoisePogs: If you forget the password, this never ends (on windows anyway). I was unable to kill the server via control+c. I think it would be better to give up after 3 wrong password attempts. After 3 wrong password attempts, the server should quit with a message.

PR #30, opened 2018-11-15
Feature/request auth method

JerryGaoRocket:
Added API to return the authentication type configured on server side. 
This is required on front-end in my local project.


1000TurquoisePogs:
Hi Jerry, can you break this up into just a PR that does the auth method? This PR seems to have work unrelated (from separate PRs)

PR #32, opened 2018-11-22
Feature/mediation layer integration #51

fkovinAtRocket:
- reorganize/update the code from an earlier attempt at integration
  - especially, restore the old plugin import resolution code in webapp
- slightly reorganize the way we handle swagger catalogs (to prepare to actually support that)

the config section now looks like this:
```
    "mediationLayer": {
      "server": {
        "hostname": "localhost",
        "port": 10011,
        "isHttps": false,
        "username": "eureka",
        "password": "password"
      },
      "enabled": true
    },
```

### DIFF
--- a/js/apiml.js
+++ b/js/apiml.js
@@ -1,0 +1,194 @@
+const Promise = require("bluebird");
+const eureka = require('eureka-js-client').Eureka;
+const zluxUtil = require('./util');
+
+const log = zluxUtil.loggers.apiml;
+
+const MEDIATION_LAYER_EUREKA_DEFAULTS = {
+  "preferSameZone": false,
+  "requestRetryDelay": 10000,
+  "heartbeatInterval": 3000,
+  "registryFetchInterval": 10000,
+  "fetchRegistry": false,
+  "availabilityZones": {
+    "defaultZone": ["defaultZone"]
+  }, 
+};
+
+
+const MEDIATION_LAYER_INSTANCE_DEFAULTS = {
+  instanceId: "localhost:zowe-zlux:8543",
+  app: "zlux",
+  hostName: "localhost",
+  ipAddr: "127.0.0.1", 
+  vipAddress: "localhost",
+  status: "UP",
+  dataCenterInfo: {
+    '@class': 'com.netflix.appinfo.InstanceInfo$DefaultDataCenterInfo',
+    name: 'MyOwn'
+  },
+  leaseInfo: {
+    durationInSecs: 10,
+    renewalIntervalInSecs: 10
+  },
+  metadata: {
+    "routed-services.1.gateway-url": "/api/v1",
+    "routed-services.1.service-url": "/ZLUX",
+    "routed-services.2.gateway-url": "/ui/v1",
+    "routed-services.2.service-url": "/ZLUX",
+    "routed-services.3.gateway-url": "/ws/v1",
+    "routed-services.3.service-url": "/ZLUX",
+    "routed-services.4.gateway-url": "/v1/api-doc",
+    "routed-services.4.service-url": "/ZLUX",
+
+    'mfaas.discovery.catalogUiTile.id': 'zlux',
+    'mfaas.discovery.catalogUiTile.title': 'ZLux Proxy Server',
+    'mfaas.discovery.catalogUiTile.description': 'The Proxy Server is an '
+       + 'HTTP, HTTPS, and Websocket server built upon NodeJS and ExpressJS. '
+       + 'This serves static content via "Plugins", and is extensible by '
+       + 'REST and Websocket "Dataservices" optionally present within Plugins.',
+    'mfaas.discovery.catalogUiTile.version': '1.0.0',
+
+    'mfaas.discovery.service.title': 'ZLux Proxy Server',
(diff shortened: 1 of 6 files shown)


fkovinAtRocket:
There should be no race conditions now.

js/webapp.js
```diff
@@ -161,6 +160,17 @@ const staticHandlers = {
         });
     });
     return r;
+  },
+  
+  eureka() {
+    const router = express.Router();
+    router.get('/application/info', function(req, res, next) {
```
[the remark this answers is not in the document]
fkovinAtRocket: It's flexible: it's actually specified by us when registering to the APIML, see here: https://github.com/zowe/zlux-proxy-server/blob/fa8cd1b5c6b63033aa33f81b9f3951aa1e7d5a72/js/apiml.js#L92

fkovinAtRocket:
Adding this makes the client more or less work:
```
diff --git a/bootstrap/src/uri/mvd-uri.ts b/bootstrap/src/uri/mvd-uri.ts
index 4d15c3b..c71358a 100644
--- a/bootstrap/src/uri/mvd-uri.ts
+++ b/bootstrap/src/uri/mvd-uri.ts
@@ -12,8 +12,8 @@
 
 import { PluginManager } from 'zlux-base/plugin-manager/plugin-manager'
 
-const proxy_path = 'zowe-zlux';
-const proxy_mode = (window.location.pathname.split('/')[1] == proxy_path) ? true : false;
+const proxy_path = 'api/v1/zlux';
+const proxy_mode = true; //(window.location.pathname.split('/')[1] == proxy_path) ? true : false;
 
 export class MvdUri implements ZLUX.UriBroker {
   rasUri(uri: string): string {
@@ -81,7 +81,14 @@
   }
 
   serverRootUri(uri: string): string {
-    return proxy_mode ? `/${proxy_path}/${uri}` : `/${uri}`;
+    if (proxy_mode) {
+      if (uri.startsWith('/')) {
+        uri = uri.substr(1);
+      }
+      return `/${proxy_path}/${uri}`;
+    } else {
+      return `/${uri}`;
+    }
   }
 
   pluginResourceUri(pluginDefinition: ZLUX.Plugin, relativePath: string): string {
```
Would like to know what a good way would be.

js/webapp.js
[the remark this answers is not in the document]
1000TurquoisePogs: Then, can it be changed to:
```
statusPageUrl: `${proto}://${this.hostName}:${port}/server/apiml/info`,
healthCheckUrl: `${proto}://${this.hostName}:${port}/server/apiml/health`,
```
Because, these are very specific pages, and /application will be confused since this is an application server.

js/apiml.js
```diff
@@ -0,0 +1,194 @@
...
+  leaseInfo: {
+    durationInSecs: 10,
+    renewalIntervalInSecs: 10
+  },
+  metadata: {
+    "routed-services.1.gateway-url": "/api/v1",
+    "routed-services.1.service-url": "/ZLUX",
+    "routed-services.2.gateway-url": "/ui/v1",
+    "routed-services.2.service-url": "/ZLUX",
+    "routed-services.3.gateway-url": "/ws/v1",
```
1000TurquoisePogs: Isn't it the case that for us, /api/v1, /ui/v1, and /ws/v1 really mean the same thing? What will happen if we don't have these as distinct routes? If possible, only 1 is needed.
Consider:
https://host/api/v1/ZLUX/plugins/foo/services/bar1/baz
wss://host/api/v1/ZLUX/plugins/foo/services/bar2/baz
2 different services, but they're both apis. They also follow the v1 structure of our server namespacing.
So, I think it makes a lot more sense to have only "routed-services.1" if possible.

js/apiml.js
```diff
@@ -0,0 +1,194 @@
...
+    "routed-services.3.service-url": "/ZLUX",
+    "routed-services.4.gateway-url": "/v1/api-doc",
+    "routed-services.4.service-url": "/ZLUX",
+
+    'mfaas.discovery.catalogUiTile.id': 'zlux',
+    'mfaas.discovery.catalogUiTile.title': 'ZLux Proxy Server',
```
1000TurquoisePogs: We really need a name change. @jhenryRocket don't we use "Zowe Application Server"? I'd also find "zlux server" to be appropriate. But, I think we need to get "proxy" out of the name because when people see Zowe, they see components and zlux's server's specialty is more about serving apps than proxying, even though it can do both.

jhenryRocket: Yes - we use "Zowe Application Server" in the user doc. I agree that we should remove the "proxy" part.

js/apiml.js
fkovinAtRocket: This is some APIML convention. A quote from Petr Plavjanik's e-mail:
> Also the prefixes /ui/v1 and /api/v1 are there to follow conventions that we have in APIML and at the end they are both mapped to the same URL in zLUX.

js/webapp.js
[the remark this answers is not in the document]
fkovinAtRocket: Well, I wouldn't have 'apiml' in those URIs though, because it's *our* health, not APIML's.

PR #34, opened 2018-11-28
Support passphrase in https options

plavjanik:
This PR adds a new field to the https options named `passphrase` 
to allow PKCS12 key store (provided in pfx option) with a password.

The certificates generated by Zowe APIML can use passwords.

Resolves zowe/api-layer#92

### DIFF
--- a/js/webserver.js
+++ b/js/webserver.js
@@ -63,6 +63,9 @@ WebServer.prototype = {
     if (this.config.https.certificateRevocationLists) {
       this.httpsOptions.crl = util.readFilesToArray(this.config.https.certificateRevocationLists);
     };
+    if (this.config.https.passphrase) {
+      this.httpsOptions.passphrase = this.config.https.passphrase
+    }
   },
 
   isConfigValid(config) {


plavjanik:
> Is this really a good idea? It ends up putting a passphrase as plaintext, right next to the pfx, making the pfx rather trivially encrypted. The zlux server has the ability to prompt for a passphrase interactively (PR for fix: #29) if a pfx is encrypted, but that wouldn't work programmatically.

In the Java world, the password/passphrase for keystore is not meant for access protection but for integrity check: 

> **-storepass[:env| :file] argument** The password that is used to protect the integrity of the keystore.
from https://docs.oracle.com/javase/8/docs/technotes/tools/windows/keytool.html

The point is that we want to start the server without without need for user interaction and the then passwords are stored in the configuration files. 

I have removed the password from the certificates generated by APIML for zLUX so this argument is not needed but I still think that it will be useful.

plavjanik:
@jackjia-ibm

PR #38, opened 2018-12-10
Check availability of proxied server

lchudinov:
Node code running on Windows should check if auth services host is configured, and issue clear instructions if not configured.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -21,7 +21,7 @@ const WebAuth = require('./webauth');
 const unp = require('./unp-constants');
 const http = require('http');
 const ApimlConnector = require('./apiml');
-const net = require('net');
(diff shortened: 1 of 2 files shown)


js/index.js
```diff
@@ -198,6 +204,27 @@ Server.prototype = {
       }
     };
     return this.webApp.installPlugin(pluginContext);
+  },
+
+  checkProxiedHost: function (host, port) {
```
fkovinAtRocket: The code looks alright but I would discourage adding stuff to index.js directly. The purpose of index.js is to do the wiring between components, and not to be a dump  for "miscellaneous" functionality, even if it looks small.

Please move this to, let's say, proxy.js

1000TurquoisePogs: That works for me too.

PR #39, opened 2018-12-13
node server tracing: login and logout (+ comments, bugfix and formatting)

fkovinAtRocket:


js/webauth.js
```diff
@@ -168,11 +170,18 @@ module.exports = function(authManager) {
...
             authServiceHandleMaps[pluginID];
-          const wasAuthenticate = authPluginSession.authenticated;
+          // FIXME This is a bug: in webauth.js we shouldn't make assumptions
+          // about the contents of the session. We only can look at the return 
+          // values of the method.
+          const wasAuthenticated = authPluginSession.authenticated;
           const handlerResult = yield handler.authenticate(req, 
               authPluginSession);
+          if (handlerResult.success) {
+            authLogger.debug(`Successful authenticate to auth handler ${pluginID}. `
```
1000TurquoisePogs: Which user logged in?

js/webauth.js
```diff
@@ -191,6 +200,7 @@ module.exports = function(authManager) {
       const handlers = getRelevantHandlers(authManager, req.body);
       for (const handler of handlers) {
         const pluginID = handler.pluginID;
+        authLogger.debug(`User logout for auth handler ${pluginID}`);
```
1000TurquoisePogs: Which user? For a given handler, there should be 1 user that is associated with a session. Since we know the handler, don't we know the user?

fkovinAtRocket:
See https://github.com/zowe/zlux-proxy-server/pull/42 for a version that adds routing logging to the dataservice versioning version

js/webauth.js
fkovinAtRocket: I think that "user" is a handler-level concept.  My idea was that we print the handler's result here: https://github.com/zowe/zlux-proxy-server/blob/cb8183d6a70127b431c4b0d5292cf11a9f53922c/js/webauth.js#L181 and currently it's just `{ success: true }`.  We could make the handler add there any information it wants to report, like the user name, if applicable.

1000TurquoisePogs:
Apparently redundant now.

PR #43, opened 2018-12-14
Support binding to individual IPs

1000TurquoisePogs:
As opposed to current behavior which is to bind HTTP & HTTPS to 0.0.0.0 on whatever IP you requested.
Utilizes the structure here: https://github.com/zowe/zlux-example-server/pull/30


1000TurquoisePogs:
Resolves https://github.com/zowe/zlux-proxy-server/issues/20

js/webserver.js
```diff
@@ -104,62 +143,94 @@ WebServer.prototype = {
...
+      if (this.config.https.ipAddresses) {
+        let httpsIps = sanitizeIps(this.config.https.ipAddresses);
+        let len = httpsIps.size;
+        if (len > 0) {
+          const iterator = httpsIps.entries();
+          for (let ipAddress of iterator) {
+            yield* makeHttpsServer(ipAddress[0], this.config.https.port);
+          }
+        } else {
+          makeHttpsServer('0.0.0.0', this.config.https.port);
```
fkovinAtRocket: missing `yield*`

js/webserver.js
```diff
@@ -104,62 +143,94 @@ WebServer.prototype = {
...
+        }
+      } else {
+        makeHttpsServer('0.0.0.0', this.config.https.port);
```
fkovinAtRocket: missing `yield*`

js/webserver.js
```diff
@@ -104,62 +143,94 @@ WebServer.prototype = {
...
     }
   }),
 
-  callListen(methodServer, methodName, methodNameForLogging) {
-    var methodConfig = this.config[methodName];
-    var addressForLogging = methodConfig.hostname ? methodConfig.hostname : "*";
-    addressForLogging += ":" + methodConfig.port;
-
+  callListen(methodServer, methodName, methodNameForLogging, ipAddress, port) {
+    var addressForLogging = `${ipAddress}:${port}`;
```
fkovinAtRocket: The use of `var` vs `let` is inconsistent in this PR. What I think makes sense is using `const` by default  and that will be like 90% of cases. Then, once you need a modifiable variable, you use `let` (so once a reader sees a `let` they know they should expect the value to be assigned more than once). And you typically don't use `var` at all, unless you want the weird 'hoisting' behaviour for some reason.

js/webserver.js
```diff
@@ -27,6 +27,18 @@ const bootstrapLogger = util.loggers.bootstrapLogger;
 const contentLogger = util.loggers.contentLogger;
 const childLogger = util.loggers.childLogger;
 
+function sanitizeIps(ipAddresses) {
+  let set = new Set();
+  ipAddresses.forEach((ipAddress)=> {
+    if (typeof ipAddress == 'string') {
+      set.add(ipAddress);
```
fkovinAtRocket: I'm wondering what we should do if we get synonymous addresses here, e.g. `[ 'localhost', '127.0.0.1' ]` or `[ 'rs22', 'rs22.rocketsoftware.com' ]`? The fact that it's a Set won't help us, and we'll try to bind to the same address twice.

rpenny125: Did you test what happens if you do that? Think what you want the outcome to be, and figure out how to achieve the outcome. I suspect you might get an error if you try to bind separately to synonyms (such as localhost and ::1 or 127.0.0.1).

If you do get errors, I suggest issuing an informative error message. As to whether you proceed or abort after that, that's a discussion.

js/webserver.js
1000TurquoisePogs: Right; sometimes this gets missed due to habit. I'm fixing this.

1000TurquoisePogs:
Please give it another check. Thanks!

1000TurquoisePogs:
I added the final missing part: that the loopback calls were unconditionally localhost... while potentially someone didn't ask for localhost in the list. This changes it to a best effort mode instead, and prints a warning if a localhost equivalent ip wasn't supplied at startup.

js/webserver.js
fkovinAtRocket: The use of Set drew my attention. It looks like we're trying to provide a feature but didn't finish the job.  If we really care to remove synonymous addresses for the user, then a Set of string hostnames won't be enough, we'd also need to resolve them and see that they're actually unique.  Or we might not care and just let the user make sure there're no synonyms. Then again we don't need a Set here.

js/webapp.js
```diff
@@ -182,15 +182,11 @@ const staticHandlers = {
...
-  if (httpsPort) {
-    this.port = httpsPort;
-    this.isHttps = true;
-  } else {
-    this.port = httpPort;
-    this.isHttps = false;
-  }
+  this.isHttps = networkConfig.isHttps;
+  this.port = networkConfig.port;
+  this.host = networkConfig.host;
```
fkovinAtRocket: Having a hostname in every service handle implies that different handles may be for different hosts, so it suggests a distributed architecture (which we don't have and don't plan). I'd rather have this as a global variable or constant.

js/webapp.js
```diff
@@ -433,6 +429,45 @@ ImportManager.prototype = {
   
 }
 
+function getLoopbackDefaults(webappOptions) {
+  //if this isn't here, throw
+  let nodeConfig = webappOptions.serverConfig.node;
+
+  let getPreferredAddress = function(addresses) {
```
fkovinAtRocket: This looks like it needs to be moved to util.js

1. It doesn't seem to refer to the local variables of the surrounding function 
2. It can be usable elsewhere
3. In this PR: https://github.com/zowe/zss-auth/pull/8/files#diff-9c4028ca3fee8475c7b0f0a4f5a92580R115 I actually used the 'ipaddr.js' library to check if an IP address is loopback, because e.g. 127.0.0.2 is also a valid loopback addr.

js/webapp.js
```diff
@@ -433,6 +429,45 @@ ImportManager.prototype = {
...
+    if (!addresses) {
+      return '127.0.0.1';
+    }
+    for (let i = 0; i < addresses.length; i++) {
+      let address = addresses[i];
+      if (address == '127.0.0.1' || address == 'localhost' || address == '0.0.0.0') {
```
fkovinAtRocket: In this PR https://github.com/zowe/zss-auth/pull/8/files#diff-9c4028ca3fee8475c7b0f0a4f5a92580R115 I used 'ipaddr.js' to check whether an address is loopback: entire ranges of loopback addresses  are defined, so, e.g. 127.0.0.2 is also valid.

js/webapp.js
1000TurquoisePogs: OK - wasn't sure where to put it so that makes sense.

js/webapp.js
1000TurquoisePogs: OK. The comments also note that this whole thing should go away if we can route directly some day.

rpenny125: While we are making network calls to ourselves, we have to make sure that we make a call to one of the addresses on which we are listening, which may not include any loopback addresses. I haven't read the code, but this discussion makes me ask the question: do we account for that?

1000TurquoisePogs: Yes. If no loopback is found, the first address is used instead.

js/webapp.js
1000TurquoisePogs: OK, well it's already a dependency from other libraries so I'll pull it in, but note that it doesn't cover all the cases here: not 0.0.0.0 and not localhost (really, no mention of it in their source)

fkovinAtRocket: I guess it's because it's an IP address manipulation library, not a hostname/DNS library.

js/webapp.js
```diff
@@ -499,6 +503,7 @@ function WebApp(options){
       maxAge: sessionTimeoutMs
     }
   }));
+  loopbackConfig = zluxUtil.getLoopbackFromConfig(options);
```
fkovinAtRocket: How does this work? loopbackConfig is a const: https://github.com/zowe/zlux-proxy-server/pull/43/files#diff-55d2935735e0c392902cca19a95e92c0R480

fkovinAtRocket:
I've implemented some general hostname resolution/unique hostname detection/better loopback address selection here: https://github.com/1000TurquoisePogs/zlux-proxy-server/pull/3

PR #49, opened 2019-01-03
Added conditionally used middleware to add no-caching headers for secure APIs

1000TurquoisePogs:
Resolves https://github.com/zowe/zlux-proxy-server/issues/36
This adds headers to all dataservice calls (unless opt-out at service level or individual call level) which instructs browsers not to cache a response. This is useful for sensitive information or calls that have no need to be cached.


1000TurquoisePogs:
Note that the "service" object passed through this section of the code is such that the properties of import services are always the original service that was imported. So, an import can not decide to have different caching behavior from the original service for example.

PR #50, opened 2019-01-04
Fix /plugins showing each plugin twice

1000TurquoisePogs:
Somehow conflict resolution resulted in this.plugins.push appearing twice in install plugin

### DIFF
--- a/js/webapp.js
+++ b/js/webapp.js
@@ -935,7 +935,6 @@ WebApp.prototype = {
       this._installPluginStaticHandlers(plugin, urlBase);      
       //import resolution will be postponed until all non-import plugins are loaded
       //only push plugin if no exceptions were seen
-      this.plugins.push(plugin);
     } catch (e) {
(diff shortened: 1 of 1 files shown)


js/webapp.js
```diff
@@ -935,7 +935,6 @@ WebApp.prototype = {
       this._installPluginStaticHandlers(plugin, urlBase);      
       //import resolution will be postponed until all non-import plugins are loaded
       //only push plugin if no exceptions were seen
```
rpenny125: I think line 937 is no longer relevant after your deletion. You can decide about line 936...

1000TurquoisePogs: I'll move it.. it's still relevant but in the wrong place by a few lines.

1000TurquoisePogs:
I'm under the assumption that this was caused from a result of online conflict resolution. It's some mystery machine, so I'll avoid doing that from now on.

PR #52, opened 2019-01-09
Bugfix for disallowing non-strings such as false to be put in as potential IPs for binding

1000TurquoisePogs:
dns lookup should throw for invalid hostnames, but for some reason does not throw when given the boolean 'false', so do a string check first.

Signed-off-by: Sean Grady <sgrady@rocketsoftware.com>


1000TurquoisePogs:
I agree and will keep that in mind

PR #53, opened 2019-01-09
Location of plugin on file system should not longer be supplied

jordanfilteau1995:
Signed-off-by: Jordan Filteau <jfilteau@rocketsoftware.com>

### DIFF
--- a/js/plugin-loader.js
+++ b/js/plugin-loader.js
@@ -231,7 +231,7 @@ Plugin.prototype = {
   },
   
   initStaticWebDependencies() {
-    if (this.contentPath) {
+    if (this.webContent) {
       let contentPath = path.join(this.location, "web");
       if (!fs.existsSync(contentPath)) {
         bootstrapLogger.warn(`plugin ${this.identifier} has web content but `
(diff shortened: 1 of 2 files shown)


jordanfilteau1995:
@1000TurquoisePogs  @fkovinAtRocket

fkovinAtRocket:
```
      let contentPath = path.join(this.location, "web");
      if (!fs.existsSync(contentPath)) {
        [...]
      } else {
        [...]
        this.pluginPath = contentPath;
      }
```
So basically this is saying `plugin.pluginPath = plugin.location + "/web"` which I find confusing: why would a plugin have a plugin path that is different from its location?  The variable name is confusing.

What problem is this solving? You're talking about exportDef() --  does this have to deal with the /plugins root service?

1000TurquoisePogs:
Fyodor - you may want to take another brief look as I just went over this and suggested and had changes pushed.

fkovinAtRocket:
Wow, it changed right as I was looking at it.

fkovinAtRocket:
OK, looks pretty safe now.

PR #57, opened 2019-01-11
App packager and installer

1000TurquoisePogs:
This is a rough draft / first step at making App management nice & easy. Following in the footsteps App packaging such as in android, Apps can be packaged using a zip-like archive, where there's a metadata file within that instructs the server as to what is going to be included (compiled output) and what can be excluded from packaging (source & dev libraries). For zlux, this is are pre-existing pluginDefinition.json structure. So, the package-app.js node script when given a folder, is able to confirm/deny that it is given the folder of a valid App, and if so, only package that which the server can consume, and leave out any source/dev tooling. NodeJS includes zlib, a library that is cross-platform for doing archiving and compression, so no native code is needed to make this archive: yazl, a common npm library, wraps node's zlib to make it easy to make an archive.

After doing a search, I found the file extension ".zapp" wasn't used, so I figured this would be a good extension name: zapp for zowe/zlux app.

After a person is given the archive, they can run unpackage-app.js which extracts using the reverse of yazl: yauzl. There's a pre-existing notion of a standard place on disk to store App files, and also a registry of the server of where these Apps are (in case they are not in the standard place). Using the knowledge of these locations, it's easy to wrap unpackage-app.js in very convenient scripts to automatically unpackage & register the app with a zowe server instance. I'm providing such scripts in a seperate PR for zlux-app-server.

What I didn't do:
1. No framework for install wizard / install-time config yet, so this "easy install" will only work for simple Apps that need & take no user config during installation. Help is welcome!
2. At some point it's likely that someone will want to make an APF-auth'd DLL for a ZSS plugin, and the zip archive format will not accommodate this permission. There will need to be some logic introduced to state that you have a ZSS DLL and yes you would / no you would not want the installer to APF auth it automatically.
3. Server needs a (big?) enhancement to be able to install Apps without a restart


utils/unpackage-app.js
```diff
@@ -0,0 +1,192 @@
...
+      + '--zluxConfig | -c ZLUXCONFIGPATH [--verbose | -v] [--overwrite | -w]';
+
+//TODO if plugins get extracted read-only, then how would we go about doing upgrades? read-write for now!
+const FILE_WRITE_MODE = 0o600;
+const DIR_WRITE_MODE = 0o700;
+
+const OPTION_ARGS = [
+  new argParser.CLIArgument('inputApp', 'i', argParser.constants.ARG_TYPE_VALUE),
+  new argParser.CLIArgument('outputDir', 'o', argParser.constants.ARG_TYPE_VALUE),
+  new argParser.CLIArgument('pluginsDir', 'p', argParser.constants.ARG_TYPE_VALUE),
```
rpenny125: It feels like one should drop the -p switch and just leave in the -c switch. Or, if the plan is to allow the person doing the install to control whether the plugin locator is added to site, instance, group or particular-user, then the switch should be about that choice and the zluxserver.json file would be used to choose the correct directory based on that choice.

1000TurquoisePogs: Just for flexibility in case the user knew the plugin dir but didn't know the location of the zlux config file, for some reason.

1000TurquoisePogs:
Closing this as premature: It's possible that we will adopt a different packaging mechanism than zips.

PR #62, opened 2019-01-18
Run unit tests via Jenkinsfile

dmitrynikolaev:
Jenkinsfile contains a multi-branch pipeline.

- the pipeline builds zlux superproject and run "npm test"
- if PR is related to other PRs, the pipeline can be re-started manually and PR numbers can be specified

Signed-off-by: Dmitry Nikolaev <dnikolaev@rocketsoftware.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,112 @@
+def customParameters = []
+for (param in ["PR_EXPLORER_SERVER_AUTH", "PR_SAMPLE_ANGULAR_APP", "PR_SAMPLE_IFRAME_APP",
+        "PR_SAMPLE_REACT_APP", "PR_TN3270_NG2", "PR_VT_NG2", "PR_ZLUX_APP_MANAGER",
+        "PR_ZLUX_APP_SERVER", "PR_ZLUX_BUILD", "PR_ZLUX_EDITOR", "PR_ZLUX_NG2", "PR_ZLUX_PLATFORM",
+        "PR_ZLUX_SHARED", "PR_ZLUX_WORKFLOW", "PR_ZLUX_WORKFLOW", "PR_ZOS_SUBSYSTEMS",
+        "PR_ZOSMF_AUTH", "PR_ZSS_AUTH"]) {
+    customParameters.push(
+        string(
+            name: param,
+            defaultValue: "",
+            description: "Pull request number",
+            trim: true
+        )
+    )
+}
+properties([
+    buildDiscarder(logRotator(numToKeepStr: "25")),
+    parameters(customParameters)
+])
+node('docker/rsqa/base') {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,112 @@
+def customParameters = []
+for (param in ["PR_EXPLORER_SERVER_AUTH", "PR_SAMPLE_ANGULAR_APP", "PR_SAMPLE_IFRAME_APP",
+        "PR_SAMPLE_REACT_APP", "PR_TN3270_NG2", "PR_VT_NG2", "PR_ZLUX_APP_MANAGER",
+        "PR_ZLUX_APP_SERVER", "PR_ZLUX_BUILD", "PR_ZLUX_EDITOR", "PR_ZLUX_NG2", "PR_ZLUX_PLATFORM",
+        "PR_ZLUX_SHARED", "PR_ZLUX_WORKFLOW", "PR_ZLUX_WORKFLOW", "PR_ZOS_SUBSYSTEMS",
+        "PR_ZOSMF_AUTH", "PR_ZSS_AUTH"]) {
```
1000TurquoisePogs: This is very explicit knowledge of repos in a repo that doesn't depend on them.

Jenkinsfile
```diff
@@ -0,0 +1,112 @@
...
+properties([
+    buildDiscarder(logRotator(numToKeepStr: "25")),
+    parameters(customParameters)
+])
+node('docker/rsqa/base') {
+    currentBuild.result = "SUCCESS"
+    try {
+        stage('Checkout') {
+            sh """
+                git clone https://github.com/zowe/zlux.git
```
1000TurquoisePogs: Likewise zlux-server-framework doesn't depend on zlux today

Jenkinsfile
```diff
@@ -0,0 +1,112 @@
...
+            }
+            dir('zlux/zlux-server-framework'){
+                checkout scm
+            }
+        }
+        stage('Build') {
+            sh """
+                cd zlux
+                # We still don't want to install ssh key into container
+                sed -i 's/git+ssh:\\/\\/git@github.com\\/zowe/https:\\/\\/github.com\\/zowe/' sample-angular-app/webClient/package.json
```
1000TurquoisePogs: And here as well

dmitrynikolaev:
Thank you for the review.

The only purpose of this script is to run server framework unit tests, so this repository is the most proper place.
However, it was really sloppy because I wasn't well aware of the dependencies chain and was trying to make sure that everything we need is in place.
It turned out that the only dependency we actually need here is **zlux-shared/src/logging** which depends on **zlux-platform**, and it seems that the only thing we can do at the moment is to hard-code them in Jenkinsfile.

I've updated the PR to reflect that.

PR #63, opened 2019-01-18
Use the zlux server TLS certificates and key for client authentication to the APIML

fkovinAtRocket:
Replaces https://github.com/zowe/zlux-server-framework/pull/59


fkovinAtRocket:
Resolves https://github.com/zowe/api-layer/issues/172

1000TurquoisePogs:
Looks like this gives the app server the ability to have its key/cert separate from what is needed to communicate with the mediation layer, so that's a good change.

PR #69, opened 2019-01-28
Feature/auth refresh

1000TurquoisePogs:
Adds the explicit ability for client to request that a session be renewed in order to not expire.
This may be an implicit feature of doing a status check ( GET to /auth), and therefore may not need its own URL... will change if needed.

Made getStatus async because it was not checking with the auth server, so it's possible that on doing getStatus that a session had expired in the past.

### DIFF
--- a/lib/webauth.js
+++ b/lib/webauth.js
@@ -81,8 +81,9 @@ AuthResponse.prototype = {
       authCategoryResult[this.keyField] = true;
     }
     //alert client of when this session expires one way or another
-    if (!handlerResult.expms) {
-      //handler may only know expiration time due to login process, as a response, or may know ahead of time due to set value or retrievable server config.
+    if (handlerResult.authenticated && !handlerResult.expms) {
+      //handler may only know expiration time due to login process, as a response,
+      //or may know ahead of time due to set value or retrievable server config.
       handlerResult.expms = handler.sessionExpirationMS;
     }
     authCategoryResult.plugins[pluginID] = handlerResult;
@@ -145,7 +146,7 @@ module.exports = function(authManager) {
       handler.addProxyAuthorizations(req1, req2Options, authPluginSession);     
     },
     
-    getStatus(req, res) {
+    getStatus: Promise.coroutine(function*(req, res) {
       const handlers = authManager.getAllHandlers();
       const result = new StatusResponse();
       for (const handler of handlers) {
@@ -153,7 +154,7 @@ module.exports = function(authManager) {
         const authPluginSession = util.getOrInit(req.session, pluginID, {});
         let status;
         try {
-          status = handler.getStatus(authPluginSession); 
+          status = yield handler.getStatus(req, authPluginSession); 
         } catch (error) {
           status = {
             error
@@ -162,7 +163,7 @@ module.exports = function(authManager) {
         result.addHandlerResult(status, handler);
       }
       res.status(200).json(result);
-    },
+    }),
 
     refreshStatus: Promise.coroutine(function*(req, res) {
       try {
@@ -178,6 +179,10 @@ module.exports = function(authManager) {
           const wasAuthenticate = authPluginSession.authenticated;
           const handlerResult = yield handler.refreshStatus(req, 
               authPluginSession);
+          if (handlerResult.success) {
+            authLogger.info(`${req.session.id}: Successful session renewal to auth ` 
+            + `handler ${pluginID}. Plugin response: ` + JSON.stringify(handlerResult));
+          }          
           //do not modify session if not authenticated or deauthenticated
           if (wasAuthenticate || authPluginSession.authenticated) {
             setAuthPluginSession(req, pluginID, authPluginSession);
--- a/lib/webauth.js
+++ b/lib/webauth.js
@@ -131,10 +131,60 @@ StatusResponse.prototype = {
   keyField: "authenticated"
 }
 
+const SESSION_ACTION_TYPE_AUTHENTICATE = 1;
+const SESSION_ACTION_TYPE_REFRESH = 2;
+
 /*
  * Assumes req.session is there and behaves as it should
  */
 module.exports = function(authManager) {
+  const _authenticateOrRefresh = Promise.coroutine(function*(req, res, type) {
+    let functionName;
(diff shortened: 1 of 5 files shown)


lib/webauth.js
```diff
@@ -157,7 +163,40 @@ module.exports = function(authManager) {
         result.addHandlerResult(status, handler);
       }
       res.status(200).json(result);
-    },
+    }),
+
+    refreshStatus: Promise.coroutine(function*(req, res) {
```
fkovinAtRocket: You did the _authenticateOrRefresh() thing in zss-auth although the authenticate and refresh logic is quite different. Here you just copied doLogin(). It and refreshStatus() differ in just a couple lines, so I would say it makes more sense to make a generic function as well.

lib/webauth.js
```diff
@@ -157,7 +163,40 @@ module.exports = function(authManager) {
...
+        const result = new LoginResult();
+        const handlers = getRelevantHandlers(authManager, req.body);
+        const authServiceHandleMaps = 
+          req[`${UNP.APP_NAME}Data`].webApp.authServiceHandleMaps;
+        for (const handler of handlers) {
+          const pluginID = handler.pluginID;
+          const authPluginSession = getAuthPluginSession(req, pluginID, {});
+          req[`${UNP.APP_NAME}Data`].plugin.services = 
+            authServiceHandleMaps[pluginID];
+          const wasAuthenticate = authPluginSession.authenticated;
```
fkovinAtRocket: You copied a bug without the FIXME comment

lib/webauth.js
```diff
@@ -140,15 +146,15 @@ module.exports = function(authManager) {
...
+    getStatus: Promise.coroutine(function*(req, res) {
       const handlers = authManager.getAllHandlers();
       const result = new StatusResponse();
       for (const handler of handlers) {
         const pluginID = handler.pluginID;
         const authPluginSession = util.getOrInit(req.session, pluginID, {});
         let status;
         try {
-          status = handler.getStatus(authPluginSession); 
+          status = yield handler.getStatus(req, authPluginSession);
```
fkovinAtRocket: So the status is updated both implicitly (when the client calls getStatus()) and explicitly when the client calls refreshStatus()? What should be the difference for the client? Why won't it, let's say, call getStatus() in both cases, if it knows that getStatus() would also update thing anyway?

However, I initially thought of getStatus() as a pure "getter" function that doesn't change any state. Why is the implicit update necessary? Can the client simply explicitly call refreshStatus() when needed?

lib/webauth.js
1000TurquoisePogs: OK, will fix

lib/webauth.js
1000TurquoisePogs: OK, we'll make getStatus() not be usable for refresh (as in, it wont send back any new expiration info as its not intended to renew or update expiration), and keep refresh as-is.

lib/webauth.js
```diff
@@ -140,15 +196,15 @@ module.exports = function(authManager) {
...
+    getStatus: Promise.coroutine(function*(req, res) {
       const handlers = authManager.getAllHandlers();
       const result = new StatusResponse();
       for (const handler of handlers) {
         const pluginID = handler.pluginID;
         const authPluginSession = util.getOrInit(req.session, pluginID, {});
         let status;
         try {
-          status = handler.getStatus(authPluginSession); 
+         status = handler.getStatus(authPluginSession);
```
fkovinAtRocket: ultra nitpicking mode on:  this space shouldn't have been removed :)

1000TurquoisePogs: OK, I'll update.

lib/webauth.js
```diff
@@ -125,10 +131,60 @@ StatusResponse.prototype = {
...
 }
 
+const SESSION_ACTION_TYPE_AUTHENTICATE = 1;
+const SESSION_ACTION_TYPE_REFRESH = 2;
+
 /*
  * Assumes req.session is there and behaves as it should
  */
 module.exports = function(authManager) {
+  const _authenticateOrRefresh = Promise.coroutine(function*(req, res, type) {
```
fkovinAtRocket: I don't see a problem just adding a method called _authenticateOrRefresh () to the  returned object.  I've been following the convention that _foo means it's private.

We'll get compile time checking for private fields/variables after TypeScript migration anyway, so I think the OO style is preferable than nested closures.

lib/webauth.js
```diff
@@ -125,10 +131,60 @@ StatusResponse.prototype = {
...
+            req[`${UNP.APP_NAME}Data`].webApp.authServiceHandleMaps;
+      for (const handler of handlers) {
+        const pluginID = handler.pluginID;
+        const authPluginSession = getAuthPluginSession(req, pluginID, {});
+        req[`${UNP.APP_NAME}Data`].plugin.services = 
+          authServiceHandleMaps[pluginID];
+        // FIXME This is a bug: in webauth.js we shouldn't make assumptions
+        // about the contents of the session. We only can look at the return 
+        // values of the method.          
+        const wasAuthenticate = authPluginSession.authenticated;
```
fkovinAtRocket: It should be `wasAuthenticated`

lib/webauth.js
1000TurquoisePogs: It's better to do this for defensive programming... typescript doesn't actually make private fields private at runtime, and I worry about non-TS code being abusive.

lib/webauth.js
1000TurquoisePogs: This area seems odd but functional. The result of getStatus returns authenticated, while other actions return success, while in this block essentially success = authenticated. Do you remember why you wanted this distinction? I fixed the conditional slightly below since it seemed to have the same issue as your FIXME comment.

fkovinAtRocket: My idea is that the contract for auth plugins is the methods and their params/return values. The contents of the session, on the other hand, is private to the plugin, and is totally opaque to the framework.

Here, the framework made assumptions about the contents of the session, based on what two existing plugins do. If this had not been fixed, all other future plugins would've been forced to follow the same convention and also have a boolean field called "authenticated" on their session, and to maintain it properly.

And BTW, even in this case, who would've checked the session expiration?

PR #70, opened 2019-02-06
Buffer.from could throw error if content odd. Use try-catch

1000TurquoisePogs:


1000TurquoisePogs:
https://github.com/zowe/zlux-server-framework/pull/71 much closer to the root cause

PR #71, opened 2019-02-06
Reject put if body is missing or not JSON

1000TurquoisePogs:
Body check for PUT was insufficient: body parser logic would set the body to an object if empty, so the existence check for a string was failing.
Now, check for a string, and check that it's JSON (also catches empty case)

### DIFF
--- a/plugins/config/lib/configService.js
+++ b/plugins/config/lib/configService.js
@@ -1499,7 +1499,7 @@ function replaceOrCreateFile(response, filename, directories, scope, relativePat
           try {
             fs.fstat(fd,(err,stats)=> {
               if (err) {
-                respondWithJsonError(response,"Failed to stat item.",HTTP_STATUS_INTERNAL_SERVER_ERROR,location);                
+                respondWithJsonError(response,"Failed to stat item.",HTTP_STATUS_INTERNAL_SERVER_ERROR,location);
               } else {
                 let maccess = -1;
                 if (stats.mtimeMs) {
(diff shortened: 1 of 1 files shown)


plugins/config/lib/configService.js
```diff
@@ -2524,37 +2524,52 @@ function ConfigService(context) {
...
-          replaceOrCreateDirectoryFiles(response,itemName,request.currentResourceObject,request.currentResourceList,request.directories,
-          request.scope,lastPath,request.resourceURL);
-        */
+
+    if (typeof request.body == 'string') {
+      try {
+        //We only support JSON storage for now.
+        //If we attempt to write out a string that isnt JSON, retrieval will be broken.
+        //This also handles the case in which body was just empty...
+        const bodyTest = JSON.parse(request.body);
```
fkovinAtRocket: Style suggestion: the catch block for this is almost 40 lines away (and it would catch and report any exception as a JSON error), and the variable is never used, so it would be puzzling to figure out why this call is here. I'd improve the "locality" of this code (NOT TESTED):

```
    if (typeof request.body !== 'string') {
      respondWithJsonError(response,"Could not access PUT body.",HTTP_STATUS_BAD_REQUEST,request.resourceURL);
      return;
    }
    try {
      JSON.parse(request.body);
    } catch (e) {
      respondWithJsonError(response,"PUT body is not JSON.",HTTP_STATUS_BAD_REQUEST,request.resourceURL);
      return;
    }
    let b64 = request.query.b64;
    ...
  ```

fkovinAtRocket:
Other than that, this looks quite safe. I don't know any context, but this shouldn't hurt.

plugins/config/lib/configService.js
1000TurquoisePogs: Thanks. I did this now.
The reason why bodyTest isn't used is that the body is written out into a file, so keeping it as a string (as it was sent) is good, but need to check that it's valid JSON first otherwise future GETs will throw exceptions either server-side or client-side.

PR #74, opened 2019-02-24
Use secure attribute for session cookie

1000TurquoisePogs:
Switches server to mode that conditionally sets secure flag on cookie… based on whether client connects from HTTP or HTTPS


fkovinAtRocket:
How I tested this:
- First connected via HTTP, logged in, then opened the HTTPS version: I was already logged in
- First connected via HTTPS, logged in, then opened the HTTP version: I was not logged in

fkovinAtRocket:
Note that switching between HTTP and HTTPS doesn't always work. You have remove the old cookies once you switched.

PR #76, opened 2019-02-25
Allow agent proxy to use HTTP or HTTPS, even with ATTLS

1000TurquoisePogs:
If you were to use ATTLS with ZSS today, you would experience connectivity issues due to ZLUX having the hardcoded assumption that agents (such as ZSS) are HTTP, never HTTPS.
This changes that, and in a way that respects the use of ATTLS: you can specify an agent to be using HTTP or HTTPS, but in HTTP you can specify attls=true to instruct ZLUX that you still really mean to be accessing the agent over HTTPS, despite what the agent itself thinks.

Example agent object in zluxserver.json:
```json
  "agent": {
    //host is for zlux to know, not zss
    "host": "localhost",
    "http": {
      "ipAddresses": ["127.0.0.1"],
      //to be a replacement for zssPort
      "port": 8542,
      "attls": true
    }
  }
```

How I tested this (without explicitly setting up ATTLS):
- try to login to desktop in standard configuration, it succeeds
- set agent.http.attls=true, or define agent.https similar to node.https
- try to login to desktop and if fails, due to HTTP being used
- set up a second zlux server in the standard configuration
- set up first server to point to second server
- now first server thinks it is talking to an HTTPS enabled ZSS (due to proxying)
- login should succeed
Of course, you can also test this with ATTLS on ZSS, I just found the above quick for my sandbox.


fkovinAtRocket:
BTW I think this deserves a unit test

PR #82, opened 2019-04-03
Compatibility with auth handlers without refresh

1000TurquoisePogs:
Bugfix: old auth handlers were throwing an exception due to missing refresh function, which was just added between 1.0.1 and 1.10. Not all auth handlers will support refresh, so expecting the function to exist is perhaps wrong. If the function doesnt exist, the success/fail of the call is now just on if getStatus's authentication status is good. /auth-refresh is best-effort, so that if it doesn't refresh, you'll still just timeout later. But for things that dont actually support refresh at all, we don't want the return to be a 400 or 500 if a subset of handlers supported & succeeded, but others simply didnt support refresh.

### DIFF
--- a/lib/webauth.js
+++ b/lib/webauth.js
@@ -159,18 +159,26 @@ module.exports = function(authManager) {
         const authPluginSession = getAuthPluginSession(req, pluginID, {});
         req[`${UNP.APP_NAME}Data`].plugin.services = 
           authServiceHandleMaps[pluginID];
-        const wasAuthenticated = handler.getStatus(authPluginSession).authenticated;
-        const handlerResult = yield handler[functionName](req,
-                                                          authPluginSession);
-        if (handlerResult.success) {
-          authLogger.info(`${req.session.id}: Session security call ${functionName} succesful for auth ` 
-                          + `handler ${pluginID}. Plugin response: ` + JSON.stringify(handlerResult));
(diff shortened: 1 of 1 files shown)


fkovinAtRocket:
This feels more like a temporary hack. If that's urgent then we might want to merge it, but ideally I think we'll need an explicit way to know if an auth type supports refresh or not. Maybe we can even deal with such plugins differently.

1000TurquoisePogs:
It is a temporary hack. I was also not aware that there were products with other auth types, so this skipped over our testing.
What would you suggest instead? 
Imagine in long term: auth handlers have a function getCapabilities that returns some number that tells you if it supports authorization, refreshing, adding proxy session tokens, or other capabilities we haven't thought about yet.
Short term problem: because we dont already have a getCapabilities function, we still need a temporary hack to assume that functions that don't have this only have authenticate() and authorize() as those are the two oldest calls that we've used.

fkovinAtRocket:
I'd say that indeed we might have a function like `getCapabilities()` that would return not a number but an object like `{ sessionRefersh: true, authorization: true }`, etc. If the handler object is missing a `getCababilities` property we just assume the default combination is in effect (no refresh).

1000TurquoisePogs:
Closing due to #118

PR #92, opened 2019-04-24
Java service support via tomcat management

1000TurquoisePogs:
First implementation of a concept by which zlux can understand and cooperate with services that are not javascript or built-in. This implementation allows services to be provided through java servlets, and handled at runtime via automatic setup of a tomcat server, and establishment of a proxy through zlux to manage it.
This introduces a new dataservice type, "java-war". Later on, the idea is to introduce new language-specific types as well as remote types to provide flexibility and choice.


DivergentEuropeans:
What a doozy @1000TurquoisePogs

PR #94, opened 2019-04-29
make extra unused auth plugins not throw errors

toddwellman:
Signed-off-by: toddwellman <twellman@rocketsoftware.com>

### DIFF
--- a/lib/plugin-loader.js
+++ b/lib/plugin-loader.js
@@ -427,6 +427,16 @@ NodeAuthenticationPlugIn.prototype = {
     }
     return true;
   },
+
+  isRequestedAuthPlugin(context) {
+    //we should not load authentication types that are 
+    //not requested by the administrator
+    if (!context.authManager.authPluginRequested(this.identifier,
+      this.authenticationCategory)) {
+      return false;
+    }
+    return true;
+  },
   
   exportDef() {
     return Object.assign(super.exportDef(), {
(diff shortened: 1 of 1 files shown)


lib/plugin-loader.js
```diff
@@ -427,6 +427,16 @@ NodeAuthenticationPlugIn.prototype = {
     }
     return true;
   },
+
+  isRequestedAuthPlugin(context) {
+    //we should not load authentication types that are 
+    //not requested by the administrator
+    if (!context.authManager.authPluginRequested(this.identifier,
```
1000TurquoisePogs: if the value of the function return is equal to the value that you are returning, then you can just return the function itself, as `return context.authManager.authPluginRequested(this.identifier, this.authenticationCategory)`, or alternatively since it is a one-liner at that point, that this can be done inline.

lib/plugin-loader.js
```diff
@@ -513,7 +503,8 @@ function makePlugin(def, pluginConfiguration, pluginContext, dynamicallyCreated)
   const self = new pluginConstr(def, pluginConfiguration);
   self.dynamicallyCreated = dynamicallyCreated;
   if (!self.isValid(pluginContext)) {
-    if (def.pluginType == 'nodeAuthentication' && !self.isRequestedAuthPlugin(pluginContext)) {
+    if (def.pluginType == 'nodeAuthentication' && pluginContext.authManager.authPluginRequested(self.identifier,
```
1000TurquoisePogs: This is consolidated, but now backwards.

PR #95, opened 2019-04-29
add port to cookie name to differentiate multiple servers on same domain

toddwellman:
Signed-off-by: Todd Wellman <twellman@rocketsoftware.com>

### DIFF
--- a/lib/webapp.js
+++ b/lib/webapp.js
@@ -530,10 +530,7 @@ function WebApp(options){
   if (options.sessionTimeoutMs) {
     sessionTimeoutMs = options.sessionTimeoutMs;
   }
-  let port = options.httpsPort;
-  if (!options.httpsPort) {
-    port = options.httpPort;
-  }
+  let port = options.httpsPort ? options.httpsPort : options.httpPort;
   this.expressApp.use(cookieParser());
   this.expressApp.use(session({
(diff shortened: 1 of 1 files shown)


lib/webapp.js
```diff
@@ -533,6 +533,7 @@ function WebApp(options){
   this.expressApp.use(cookieParser());
   this.expressApp.use(session({
     //TODO properly generate this secret
+    name: 'connect.sid.' + options.httpsPort,
```
1000TurquoisePogs: This is not 100%.... it's possible for someone to have enabled http and disabled https (though discouraged).
Instead, check if https ISNT being used. If it isn't, use http. If both are being used, you can still use https since you have to choose only 1.

lib/webapp.js
```diff
@@ -530,9 +530,14 @@ function WebApp(options){
   if (options.sessionTimeoutMs) {
     sessionTimeoutMs = options.sessionTimeoutMs;
   }
+  let port = options.httpsPort;
+  if (!options.httpsPort) {
+    port = options.httpPort;
+  }
```
1000TurquoisePogs: Better way to do this is:
`const port = options.httpsPort ? options.httpsPort : options.httpPort;`
Allows use of const for read-only, and also does not need reassignment.

PR #97, opened 2019-05-08
Remove cookie exp & add server exp

1000TurquoisePogs:
The zlux cookie used to have a timeout on it, but it didn't make too much sense versus the individual auth timeouts.
Now, this has been removed in favor of server-side tracking, and the server counts an expiration based on the latest timeout for all auths related to a session.
This makes use of the session object that comes from the cluster, so this works both in single process mode and in cluster mode.

### DIFF
--- a/lib/auth-manager.js
+++ b/lib/auth-manager.js
@@ -88,10 +88,11 @@ AuthManager.prototype = {
     let plugin;
     while ((plugin = this.pendingPlugins.pop()) !== undefined) {
       try {
-        const authenticationHandler = yield plugin.authenticationModule(plugin,
-                                                                        this.configuration,
-                                                                        config,
-                                                                        new AuthPluginContext(plugin));
+        const authenticationHandler = yield plugin.authenticationModule(
+                                              plugin,
+                                              this.configuration,
+                                              config,
+                                              new AuthPluginContext(plugin));
         // at this time we should have resolved plugin configuration to have a 
         // nice list of info about what we are using to authenticate against
         if ((typeof authenticationHandler.authenticate) !== 'function') {
--- a/lib/auth-manager.js
+++ b/lib/auth-manager.js
@@ -19,9 +19,17 @@ const zluxUtil = require('./util');
 const bootstrapLogger = zluxUtil.loggers.bootstrapLogger;
 const authLog = zluxUtil.loggers.authLogger;
 
+const DEFAULT_SESSION_TIMEOUT_MS = 60 /* min */ * 60 * 1000;
+
 const acceptAllHandler = {
   authorized() {
     return Promise.resolve({ authorized: true });
+  }
+}
+
+class AuthPluginContext {
+  constructor(plugin) {
+    this.logger = global.COM_RS_COMMON_LOGGER.makeComponentLogger(plugin.identifier);
   }
 }
 
@@ -42,6 +50,9 @@ function AuthManager(options) {
     pendingPlugins: [],
     rbacEnabled: !!options.config.rbac
   });
+  if (!this.sessionTimeoutMs) {
+    this.sessionTimeoutMs = DEFAULT_SESSION_TIMEOUT_MS;
+  }
   if (!this.rbacEnabled) {
     bootstrapLogger.warn('RBAC is disabled in the configuration. All authenticated'
         + ' users will have access to all servces. Enable RBAC in the configuration'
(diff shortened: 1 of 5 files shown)


lib/auth-manager.js
```diff
@@ -78,7 +89,9 @@ AuthManager.prototype = {
     while ((plugin = this.pendingPlugins.pop()) !== undefined) {
       try {
         const authenticationHandler = yield plugin.authenticationModule(plugin,
-                this.configuration, config);
+                                                                        this.configuration,
+                                                                        config,
+                                                                        new AuthPluginContext(plugin));
```
fkovinAtRocket: Formatting question: do we like this style? To me it looks like there's a huge void where the code should be, and then a bunch of code pieces hanging somewhere on the right.
```
foo(a,
<+4 spaces>b,
<+4 spaces>c);
```
or 
```
foo(
<+4 spaces>a,
<+4 spaces>b,
<+4 spaces>c);
```
look less like ASCII art to me.

lib/webauth.js
```diff
@@ -86,14 +92,19 @@ AuthResponse.prototype = {
       //or may know ahead of time due to set value or retrievable server config.
       handlerResult.expms = handler.sessionExpirationMS;
     }
+    //overall expiration when last auth expires
```
fkovinAtRocket: so when handler A has expired but plugin B hasn't yet, then functions protected by A will stop working, but those protected  by B wont?

fkovinAtRocket:
Just wondering: can we have unit tests for this?

lib/sessionStore.js
```diff
@@ -136,15 +136,17 @@ SessionStore.prototype.destroy = function(sid, callback) {
     try {
         //console.log('SessionStore.destroy ' + sid);
         this.removeSid(sid);
-        callback(null);
+        callback && callback(null);
```
fkovinAtRocket: so the callback wasn't optional, but now it is, right?

lib/auth-manager.js
1000TurquoisePogs: I like either, as long as observing the 120 char limit, otherwise your suggested style solves it.

lib/sessionStore.js
1000TurquoisePogs: Yes - apparently nobody was calling it at all before.

lib/webauth.js
1000TurquoisePogs: Yes. So, we're leaving it up to the UI to make good UI alerts about expiration. I just wanted an overall expiration so that the session wouldn't exist forever.

lib/webauth.js
```diff
@@ -39,10 +40,15 @@ function getAuthPluginSession(req, pluginID, dflt) {
 
 function setAuthPluginSession(req, pluginID, authPluginSession) {
   if (req.session) {
-    // FIXME Note that it does something only when req.session[pluginID] is 
+    // FIXME Note that it does something only when req.session.authPlugins[pluginID] is
```
fkovinAtRocket: so I used to simply divide the session by having `session["com.example.plugin1"]`, `session["com.example.plugin2"]`, etc, where each of those objects is the plugin's private storage. Now we have `session.authPlugins["com.example.plugin1"]` instead because we also have `session.zluxExp` right?

I'd say that, first, we might want to group all framework data also under a separate object in the session, and second, maybe that name can be a little more descriptive, like, `session.zlux.expirationTime`?

1000TurquoisePogs: zluxExp, but also because session.id existed, so previously it was technically true that a plugin could not be named "id". Not sure what wins the race. Sure, we can do zlux & zlux.expirationTime.

lib/webauth.js
```diff
@@ -172,10 +200,15 @@ module.exports = function(authManager) {
         }
         result.addHandlerResult(handlerResult, handler);
       }
-      result.updateStatus();
+      
+      result.updateStatus(authManager.sessionTimeoutMs);
+      if (result.expms !== -1) {
```
fkovinAtRocket: is this `-1` some special magic value? where is it supposed to come from?

1000TurquoisePogs: I will add a comment: -1 means they explicitly want there to be no expiration.

fkovinAtRocket: Maybe we need to also define a constant for it...

1000TurquoisePogs:
Addressed Fyodor's points, and am told Dmitry gave this a test, so I'm putting it into staging and we can continue from there. Will see if I can get a unit test into the codebase.

PR #99, opened 2019-05-15
Feature/host swagger docs from plugins

toddwellman:


lib/swagger-catalog.js
```diff
@@ -25,57 +33,134 @@ function getServiceSummary(service) {
...
     }
   }
-  return openApi;
+  return allServiceDocs;
+}
+
+function readSingleSwaggerFile (dirName, fileName) {
+  // read one swagger file and validate the json that is returned
+  return new Promise ((resolve, reject) => {
+    let fileContent = fs.readFileSync(path.join(dirName, fileName), 'utf-8');
```
1000TurquoisePogs: Promise allows asynchrony, so readFileSync just slows everything down. Please do one or the other, don't mix them. Only do sync if there is some timing issue, and even then it is best to plan a fix.

lib/swagger-catalog.js
```diff
@@ -25,57 +33,134 @@ function getServiceSummary(service) {
...
+    });
+  })
+}
+
+async function getSwaggerDocs(plugin, productCode, nodeContext) {
+  var allServiceDocs = [];
+  if (plugin.dataServices){
+    for (let i = 0; i < plugin.dataServices.length; i++) {
+      let service = plugin.dataServices[i]
+      if (service.swaggerdoc) {
```
1000TurquoisePogs: This complicates things. It's easier to enforce a naming convention: in /doc/swagger a file with the servicename + .json or .yaml should be found if present. If an import dataservice, routers should be attached appropriately later and file reading skipped.

package.json
```diff
@@ -30,8 +30,10 @@
     "express-ws": "~4.0.0",
     "glob": "~7.1.3",
     "ipaddr.js": "~1.8.0",
+    "js-yaml": "^3.13.1",
```
1000TurquoisePogs: ~, never ^

lib/swagger-catalog.js
```diff
@@ -25,57 +33,134 @@ function getServiceSummary(service) {
...
-      post: {
-        responses: {
-          200: {
-            description: "service call succeeded"
-          }
-        }
+function makeCatalogForPlugin(plugin, productCode, nodeContext) {
+  return new Promise((resolve) => {
+    const openApi = {
+      openapi: "3.0.0",
```
1000TurquoisePogs: we must use 2 not 3

PR #108, opened 2019-05-30
Exception fix for non-JSON terminal message

1000TurquoisePogs:
All terminal messages should be JSON, but in case they arent, close the connection instead of fail


1000TurquoisePogs:
It's cleaner this way since it only goes to that one place. I was actually looking for if we could try-catch around an entire express router to prevent disastrous exceptions entirely, but this seems to not be possible sadly - we just have to rely on the resilience of the nodejs cluster mode.

PR #113, opened 2019-06-18
script for changing NODE_PATH

aarora4:
Appending to node_path


1000TurquoisePogs:
We're about to make another one to fix DCO

PR #118, opened 2019-06-26
Compatibility for auth handlers without refresh (Revisited)

timgerstel:
Old authentication handlers which do not support session refreshing will not longer return a 400 or 500 error due to lack of capability, but will allow the session to expire naturally (user logs out/timeout).

Must be merged with:
https://github.com/zowe/zss-auth/pull/19
https://github.com/zowe/zosmf-auth/pull/14

Signed-off-by: Timothy Gerstel tgerstel@rocketsoftware.com

### DIFF
--- a/lib/webauth.js
+++ b/lib/webauth.js
@@ -155,15 +155,6 @@ StatusResponse.prototype = {
   keyField: "authenticated"
 }
 
-function getCapabilities(handler){
-  const handlerFuncs = ['getStatus', 'authenticate', 'refreshStatus', 'authorized', 'addProxyAuthorizations'];
-  let capabilities = {};
-  for(let func of handlerFuncs){
-    capabilities[func] = (typeof handler[func] === 'function');
-  }
-  return capabilities;
-}
-
 const SESSION_ACTION_TYPE_AUTHENTICATE = 1;
 const SESSION_ACTION_TYPE_REFRESH = 2;
 
--- a/lib/webauth.js
+++ b/lib/webauth.js
@@ -208,13 +208,16 @@ module.exports = function(authManager) {
           }
         } else {
           authLogger.warn(`${pluginID} does not have getCapabilities() function defined`);
-          handlerReuslt = (functionName == 'refreshStatus') ? { success: true } : yield handler[functionName](req, authPluginSession);
+          handlerResult = (functionName == 'refreshStatus') ? { success: true } : yield handler[functionName](req, authPluginSession);
         }
           
         if (handlerResult.success) {
           authLogger.info(`${req.session.id}: Session security call ${functionName} succesful for auth ` 
                           + `handler ${pluginID}. Plugin response: ` + JSON.stringify(handlerResult));
-        }          
+        } else {
+          authLogger.info(`${req.session.id}: Session security call ${functionName} failed for auth ` 
+                          + `handler ${pluginID}. Plugin response: ` + JSON.stringify(handlerResult));
+        }
         //do not modify session if not authenticated or deauthenticated
         if (wasAuthenticated || handlerResult.success) {
           setAuthPluginSession(req, pluginID, authPluginSession);
--- a/lib/webauth.js
+++ b/lib/webauth.js
@@ -180,23 +180,37 @@ module.exports = function(authManager) {
             req[`${UNP.APP_NAME}Data`].webApp.authServiceHandleMaps;
 
       const timeout = req.session.zlux ? req.session.zlux.expirationTime : 0;
+      //let refreshHandlers = [];
       for (const handler of handlers) {
         const pluginID = handler.pluginID;
         const authPluginSession = getAuthPluginSession(req, pluginID, {});
         req[`${UNP.APP_NAME}Data`].plugin.services = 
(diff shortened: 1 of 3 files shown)


timgerstel:
You are correct, good catch.  Will fix then comment back here, thank you @jordanfilteau1995

timgerstel:
Added fix @jordanfilteau1995

lib/webauth.js
```diff
@@ -185,24 +185,39 @@ module.exports = function(authManager) {
...
+          The logic for performing refreshStatus()/authenticate() calls has been wrapped in a check for the presence of
+          the getCapabilities() function because functions that do not include getCapabilities() likely dont support refreshing (or haven't been maintained).  
+          If a service calls refreshStatus on an auth handler that does not support it,
+          { success: true } is returned so that the session may expire naturally and users are not forced to log in every time they refresh.
+        */
+        if(hasGetCapabilities){
+          if (type === SESSION_ACTION_TYPE_REFRESH
+              && authManager.sessionTimeoutMs !== TIMEOUT_VALUE_NO_EXPIRE
+              && (!timeout || timeout < Date.now())) {
+            req.session.zlux = undefined;
```
1000TurquoisePogs: This should not be inside of hasGetCapabilities - this is checking if the server itself has timed out, not if an auth plugin has.

Consider: 
if (this stuff) { 
REASON_ZLUX_SESSION_EXPIRE
} else if (hasGetCapabilities && canRefresh) { 
do refresh
} else {
success = true
}

lib/webauth.js
```diff
@@ -185,24 +185,40 @@ module.exports = function(authManager) {
...
             && authManager.sessionTimeoutMs !== TIMEOUT_VALUE_NO_EXPIRE
             && (!timeout || timeout < Date.now())) {
           req.session.zlux = undefined;
           handlerResult = {success:false, reason:REASON_ZLUX_SESSION_EXPIRE};
+        } else if(hasGetCapabilities && handler.getCapabilities().canRefresh === true && functionName == 'refreshStatus'){
+          handlerResult = yield handler[functionName](req, authPluginSession);
         } else {
-          handlerResult = yield handler[functionName](req,
-                                                      authPluginSession);
+          handlerResult = (functionName == 'refreshStatus') ? { success: true } : yield handler[functionName](req, authPluginSession);
```
fkovinAtRocket: Line longer than 120 character

lib/webauth.js
```diff
@@ -185,24 +185,40 @@ module.exports = function(authManager) {
...
+       if(!hasGetCapabilities){
+         authLogger.warn(`${pluginID}: getCapabilities() is not a function`);
+       }
+       
+       if (type === SESSION_ACTION_TYPE_REFRESH
             && authManager.sessionTimeoutMs !== TIMEOUT_VALUE_NO_EXPIRE
             && (!timeout || timeout < Date.now())) {
           req.session.zlux = undefined;
           handlerResult = {success:false, reason:REASON_ZLUX_SESSION_EXPIRE};
+        } else if(hasGetCapabilities && handler.getCapabilities().canRefresh === true && functionName == 'refreshStatus'){
```
fkovinAtRocket: Line longer than 120 character

timgerstel: Resolved @fkovinAtRocket

lib/webauth.js
```diff
@@ -185,24 +185,38 @@ module.exports = function(authManager) {
...
+        const authHandlerStatus = handler.getStatus(authPluginSession);
+        const wasAuthenticated = authHandlerStatus.authenticated;
+        const hasGetCapabilities = !!((typeof handler.getCapabilities) === 'function');
+        let handlerResult;
+
+       if(!hasGetCapabilities){
+         authLogger.warn(`${pluginID}: getCapabilities() is not a function`);
+       }
+
+       if (type === SESSION_ACTION_TYPE_REFRESH
```
fkovinAtRocket: This is technically correct, and also probably  this is an attempt to make a minimal change, but in order to make the intent clearer, I would make a bigger change:
1. Get rid of the `functionName` variable completely from this function: we only need `type` to track what's going on, no need to use two variables for that
2. rewrite it somehow like this: (NOT TESTED!!! review carefully)
```
       if (type === SESSION_ACTION_TYPE_REFRESH) {
         if (authManager.sessionTimeoutMs !== TIMEOUT_VALUE_NO_EXPIRE
            && (!timeout || timeout < Date.now())) {
           req.session.zlux = undefined;
           handlerResult = {success:false, reason:REASON_ZLUX_SESSION_EXPIRE};
         } else if (hasGetCapabilities && handler.getCapabilities().canRefresh) {
           handlerResult = yield handler.refreshStatus(req, authPluginSession);
         } else {
           handlerResult = { success: true };
         }        
       } else if (type === SESSION_ACTION_TYPE_AUTHENTICATE) {
         handlerResult = yield handler.authenticate(req, authPluginSession);
       } 
```
In my view, this is more straightforwardly presenting what we want to do here.

timgerstel: Reviewed, implemented and tested.  Works perfectly @fkovinAtRocket

PR #122, opened 2019-06-28
Bugfix - Fix PFX Prompt & Hide Password

DivergentEuropeans:
• Fixes previous PFX syntax errors so that PFX can be used as an authentication container for Zowe keys & certificates
• Enhancement to PFX password prompt that hides password, even when terminals don't allow functionality to hide input, usually meaning the input is cooked and rawMode cannot be set (for ex: Git Bash only processes input after a line break and has no detection of keypress to mututate chars into '*' for example). This PR works by looping through and repeatedly clearing the line until the password is verified.
• As a byproduct, fixes a bug where old PFX password hiding logic (asterisks) would break with backspace on Windows

image

**Test strategy**
In Git Bash, you can generate a PFX very simply by navigating to `/zlux-app-server/deploy/product/ZLUX/serverConfig` and running
`certutil -mergepfx zlux.keystore.cer out.pfx`
Certutil should be a built-in utility that will take the zlux.keystore certificate and make it into a PFX "out.pfx"

To use the PFX, go to `zlux-app-server\deploy\instance\ZLUX\serverConfig\zluxserver.json`
and comment out

"keys": ["../deploy/product/ZLUX/serverConfig/zlux.keystore.key"],
 "certificates": ["../deploy/product/ZLUX/serverConfig/zlux.keystore.cer"],
"certificateAuthorities": ["../deploy/product/ZLUX/serverConfig/apiml-localca.cer"]

into

"pfx": "../deploy/product/ZLUX/serverConfig/out.pfx"
// "keys": ["../deploy/product/ZLUX/serverConfig/zlux.keystore.key"],
// "certificates": ["../deploy/product/ZLUX/serverConfig/zlux.keystore.cer"],
// "certificateAuthorities": ["../deploy/product/ZLUX/serverConfig/apiml-localca.cer"]

**Note: To have the password prompt appear, you need to disable output to log.** 
Currently, we don't have a way to use the terminal for output & log at the same time. Technically, the password prompt is **not** necessary for this PR to work (you could type your password without getting the output from the terminal to tell you to type a password). Anyway, this PR works with the regular old nodeServer.bat as far as I am aware, but it just doesn't show the password prompt. To have the password prompt appear in terminal, use https://github.com/zowe/zlux-app-server/pull/62

or you can go inside `C:\Users\lastrakou\Desktop\src3\zlux\zlux-app-server\bin\`
and change
 `node --harmony zluxServer.js --config=../deploy/instance/ZLUX/serverConfig/zluxserver.json %* > %ZLUX_NODE_LOG_DIR%\nodeServer.log 2>&1`
into 
`node --harmony zluxServer.js --config=../deploy/instance/ZLUX/serverConfig/zluxserver.json %*`

### DIFF
--- a/js/encryption.js
+++ b/js/encryption.js
@@ -19,6 +19,11 @@ function encryptWithKey(text,key) {
   return encrypted;
 }
 
+function getKeyFromPasswordSync(password,salt,length) {
+  var rounds = 500;
+  return crypto.pbkdf2Sync(password,salt,rounds,length,'sha256');
+}
+
 function getKeyFromPassword(password,salt,length,callback) {
   var rounds = 500;
   crypto.pbkdf2(password,salt,rounds,length,'sha256',(error, derivedKey) => {
@@ -54,6 +59,7 @@ function decryptWithKeyAndIV(text,key,iv) {
 exports.encryptWithKeyAndIV = encryptWithKeyAndIV;
 exports.decryptWithKeyAndIV = decryptWithKeyAndIV;
 exports.getKeyFromPassword = getKeyFromPassword;
+exports.getKeyFromPasswordSync = getKeyFromPasswordSync;
 exports.encryptWithKey = encryptWithKey;
 exports.decryptWithKey = decryptWithKey;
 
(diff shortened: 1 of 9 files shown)


DivergentEuropeans:
This PR solves a bug that's been around for over 6 months and the solution is semi hacky (as opposed to using already presented solutions like readline-sync for example or other sync read stream methods) because of terminals like Git Bash misbehaving and not working well with these other ways...

PR #129, opened 2019-08-30
Retreive plugin default config from plugin dir

1000TurquoisePogs:
To solve the issue of how plugins can ship default behavior that can be overridden while having read-only directories for server & plugins, plugins must keep their default config files within their root folder. So, the config service must be made to locate the plugin directory to read there as part of its decision making as to what config gets returned. This turns **plugin** into the new broadest scope of config service, but otherwise has the characteristics of the **product** scope.

### DIFF
--- a/lib/plugin-loader.js
+++ b/lib/plugin-loader.js
@@ -163,7 +163,7 @@ ExternalService.prototype = {
 }
 
 function makeDataService(def, plugin, context) {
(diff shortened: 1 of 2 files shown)


1000TurquoisePogs:
Reviewed this with @timgerstel for double-checking and understanding. No other issues noted.

PR #130, opened 2019-08-30
Support binary mode for config

1000TurquoisePogs:
Enhances the config service to support working with files other than JSON.
Non-JSON are treated as opaque files, where attribute merging will not be performed.
But, you can for example get & put JPEGs.
This requires a "resource" to be marked as binary, rather than JSON, so that there is no mix-and-match.
Some config service modes work on multiple JSON which could be aggregated into a single response. This doesn't make sense for binary, so some of these modes are not supported.

Binary files could be much larger than JSON. And expressjs limits the body size of a PUT/POST message to 100kb by default. So, for binary, this is put at 3mb at first, and we can change this number later on as the need arises.

This changes the parsing of the body slightly in that Content-Type text/plain will now be interpreted as raw rather than string, so JSON must explicitly be application/json instead.

### DIFF
--- a/plugins/config/lib/configService.js
+++ b/plugins/config/lib/configService.js
@@ -49,10 +49,11 @@ const CONFIG_SCOPE_PLUGIN = 6;
 const PERMISSION_DEFAULT_FORBID = 0;
 const PERMISSION_DEFAULT_ALLOW = 1;
 
-const CURRENT_JSON_VERSION = "0.9.0";
-
+const CURRENT_JSON_VERSION = "0.10.0";
 
 const PLUGIN_DEFAULT_DIR = pathModule.join('config','storageDefaults');
+const BINARY_BODY_SIZE_LIMIT = '3mb';
+
 //a file
 const MSG_TYPE_RESOURCE = "org.zowe.configjs.resource";
 //contents of a folder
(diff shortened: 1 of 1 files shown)


plugins/config/lib/configService.js
```diff
@@ -2046,7 +2147,13 @@ function addJSONFilesToJSON(startingPath,json) {
     fileNames.forEach(function (filename) {
```
1000TurquoisePogs: function should not be called addJSONFilestoJSON anymore, but just "addResource"... because binary is now an option.

plugins/config/lib/configService.js
```diff
@@ -2265,8 +2372,10 @@ function ConfigService(context) {
     next();
   });
 
+  
+
+  router.use(bodyParser.raw({type: binaryTypeCheck, limit:'3mb'}));
```
1000TurquoisePogs: Should add comment describing whats going on here, especially about the default value of 100kb

1000TurquoisePogs: Should put '3mb' as a constant at the top of the file

1000TurquoisePogs:
Reviewed with @timgerstel and addressed all feedback.

PR #137, opened 2019-09-12
Control plugin and services lifecycle without restarting Node.js

timgerstel:
The purpose of this epic is to provide fundamental management capabilities for the App Server so Node.js restart is no longer required to make configuration or deployment changes effective.

Node.js restart is not required when:
    adding a plugin or service 
    removing a plugin or service 
    upgrading a plugin or service 
    rewriting config (including updating port)

Do not merge without:
https://github.com/zowe/zlux-app-server/pull/69


1000TurquoisePogs:
> 
> 
> These changes are pretty severe and add huge enhancements to the server. So I think it would be good to change https://github.com/zowe/zlux-server-framework/blob/master/package.json as well to reflect that by
> 
>     1. bumping the minor version to 0.1.0 (@1000TurquoisePogs thoughts?)
> 
>     2. add @DivergentEuropeans & @timgerstel to the Contributors array

Feel free to add yourselves to the contributors array, but it isn't really seen since we dont host it on NPM yet.

PR #152, opened 2019-10-04
Cookie & WS changes

1000TurquoisePogs:
- Pass through express router ws patcher in case it is needed by plugins (on my system, for some reason router.ws was missing on one plugin)
- Allow security plugins to manage headers that are proxied, so that things not needed by the browser are not put in
- Clear cookie on complete logout

### DIFF
--- a/lib/webauth.js
+++ b/lib/webauth.js
@@ -161,7 +161,7 @@ const SESSION_ACTION_TYPE_REFRESH = 2;
 /*
  * Assumes req.session is there and behaves as it should
  */
-module.exports = function(authManager) {
+module.exports = function(authManager, cookiePort, isSecurePort) {
   const _authenticateOrRefresh = Promise.coroutine(function*(req, res, type) {
     let functionName;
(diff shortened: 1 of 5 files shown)


lib/webauth.js
```diff
@@ -246,8 +246,26 @@ module.exports = function(authManager) {
...
+        authLogger.warn(`Failed to set proxy authorizations. Error=`,e);
+      }
     },
+
+    processProxiedHeaders(req, headers) {
+      const handler = getAuthHandler(req, authManager);
+      if (handler && handler.processProxiedHeaders) {
+        const authPluginSession = getAuthPluginSession(req, handler.pluginID, {});
+        try {
+          return handler.processProxiedHeaders(req, headers, authPluginSession);
```
timgerstel: If someone were to use something other than zss-auth, the bug would still be present since the headers are returned in the case of an exception.  Should this function perhaps also be added to zosmf-auth as well?

1000TurquoisePogs: Possibly, yes. Need to check on that next.

PR #153, opened 2019-10-08
Added support for allowed plugins for certain users when rbac is enabled

sakeerthy:

### DIFF
--- a/lib/webapp.js
+++ b/lib/webapp.js
@@ -218,7 +218,7 @@ function setAttrib(object, path, value){
 }
 
 const staticHandlers = {
-  plugins: function(plugins) {
+  plugins: function(plugins, options) {
     return function(req, res) {
       let parsedRequest = url.parse(req.url, true);
       if (!parsedRequest.query) {
@@ -233,21 +233,40 @@ const staticHandlers = {
         but this should be enforced somehow in the future.
       */
       if (!type) {
-        do404(req.url, res, "A plugin type must be specified");
-        return;
+        type = "all"
       }
       const acceptLanguage = 
         translationUtils.getAcceptLanguageFromCookies(req.cookies) || req.headers['accept-language'] || '';
       const pluginDefs = plugins.map(p => p.exportTranslatedDef(acceptLanguage));
+            
+      let filteredPluginDefs = []
+      let allowedPlugins;
+
+      if (options.serverConfig.dataserviceAuthentication.rbac) {
+        if (req.username) {
+          allowedPlugins = configService.getAllowedPlugins(configService.CONFIG_SCOPE_USER, options, req.username);
+        } else {
+          allowedPlugins = configService.getAllowedPlugins(configService.CONFIG_SCOPE_PRODUCT, options);
+        }
+        for(let plugin of pluginDefs) {
+          let obj = allowedPlugins.allowedPlugins.find(o => o.identifier === plugin.identifier)
+          if (obj) {
+            filteredPluginDefs.push(plugin)
+          }
+        }
+      } else {
+        filteredPluginDefs = pluginDefs;
+      }
+
       const response = {
         //TODO type/version
         pluginDefinitions: null 
       };
       contentLogger.debug('Type requested ='+type);
       if (type == "all") {
-        response.pluginDefinitions = pluginDefs;
+        response.pluginDefinitions = filteredPluginDefs;
       } else {
-        response.pluginDefinitions = pluginDefs.filter(def => {
+        response.pluginDefinitions = filteredPluginDefs.filter(def => {
           if (def.pluginType != null) {
             contentLogger.debug('Returning true if type matches, type='
                 + def.pluginType);
@@ -1049,7 +1068,7 @@ WebApp.prototype = {
         this.appData));
   },
 
-  _installRootService(url, method, handler, {needJson, needAuth, isPseudoSso}) {
+  _installRootService(url, method, handler, {needJson, needAuth, authType, isPseudoSso}) {
     const handlers = [commonMiddleware.logRootServiceCall(false, url), commonMiddleware.httpNoCacheHeaders()];
     if (needJson) {
       handlers.push(jsonParser);
@@ -1063,7 +1082,11 @@ WebApp.prototype = {
       })
     }
     if (needAuth) {
-      handlers.push(this.auth.middleware); 
+      if (authType == "semi") {
+        handlers.push(this.auth.semiAuthenticatedMiddleware)
+      } else {
+        handlers.push(this.auth.middleware); 
+      }
     }
     handlers.push(handler);
     installLog.info(`installing root service at ${url}`);
@@ -1106,8 +1129,8 @@ WebApp.prototype = {
     this._installRootService('/auth-logout', 'get', this.auth.doLogout, 
         {needJson: true, needAuth: false, isPseudoSso: true});
     serviceHandleMap['auth'] = new WebServiceHandle('/auth', this.wsEnvironment);
-    this._installRootService('/plugins', 'get', staticHandlers.plugins(this.plugins), 
-        {needJson: false, needAuth: false, isPseudoSso: false}); 
+    this._installRootService('/plugins', 'get', staticHandlers.plugins(this.plugins, this.options), 
+        {needJson: false, needAuth: true, authType: "semi", isPseudoSso: false}); 
     this._installRootService('/plugins', 'use', staticHandlers.pluginLifecycle(this.options, this.plugins),
       {needJson: true, needAuth: true, isPseudoSso: false});
     serviceHandleMap['plugins'] = new WebServiceHandle('/plugins', this.wsEnvironment);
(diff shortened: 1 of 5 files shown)


lib/webapp.js
```diff
@@ -233,21 +233,40 @@ const staticHandlers = {
...
       const pluginDefs = plugins.map(p => p.exportTranslatedDef(acceptLanguage));
+            
+      let filteredPluginDefs = []
+      let allowedPlugins;
+
+      if (options.serverConfig.dataserviceAuthentication.rbac) {
+        if (req.username) {
+          allowedPlugins = configService.getAllowedPlugins(configService.CONFIG_SCOPE_USER, options, req.username);
+        } else {
+          allowedPlugins = configService.getAllowedPlugins(configService.CONFIG_SCOPE_PRODUCT, options);
```
1000TurquoisePogs: Not product, try CONFIG_SCOPE_INSTANCE, because thats the lowest level that doesnt require a username.

lib/webapp.js
```diff
@@ -233,21 +233,40 @@ const staticHandlers = {
...
+        }
+        for(let plugin of pluginDefs) {
+          let obj = allowedPlugins.allowedPlugins.find(o => o.identifier === plugin.identifier)
```
1000TurquoisePogs: It's possible that the above configService returns allowedPlugins = null.
If so, you should return the unfiltered pluginDefs object.

1000TurquoisePogs: Also, your filter ignores the plugin version, which is in the JSON you get.
So, right now you are not allowed to have more than one version of the same plugin in the server, but you should at least add this logic so that we dont forget to do it later... so, if it is "*", then it always succeeds, but otherwise... try the 'semver' library we already have in this codebase... see here: https://github.com/zowe/zlux-server-framework/blob/master/lib/depgraph.js#L294

lib/webauth.js
```diff
@@ -295,6 +295,37 @@ module.exports = function(authManager) {
...
+        }
+        const authPluginID = handler.pluginID;
+        let result;
+        const timeout = req.session.zlux ? req.session.zlux.expirationTime : 0;
+        if (authManager.sessionTimeoutMs !== TIMEOUT_VALUE_NO_EXPIRE
+            && (!timeout || timeout < Date.now())) {
+              req.session.zlux = undefined;
+              result = {authenticated: false, authorized: false};
+        } else {
+          const authPluginSession = getAuthPluginSession(req, authPluginID, {});
```
1000TurquoisePogs: No else here, this does authorize() and we don't care for semi-auth.
In the IF above, it should do a `res.status(401).json(result); return;`
Otherwise, `next(); return;` means success

plugins/config/lib/configService.js
```diff
@@ -767,6 +775,36 @@ function overrideJsonProperties(originalObject, overrideObject) {
...
+  for (let i = 0; i < keyArray.length; i++) {
+    key = keyArray[i];
+    property = originalObject[key];
+    overrideProperty = htGet(overrideTable,key);
+    if (!overrideProperty) {
+      overrideObject[key] = property;
+    } else {
+      if (Array.isArray(overrideProperty)) {
+        if (Array.isArray(property)) {
+          Array.prototype.push.apply(overrideProperty, property);
```
1000TurquoisePogs: Even if this works, `overrideProperty.push.apply(overrideProperty,property)` is a little bit more readable to me. IMO.

plugins/config/lib/configService.js
```diff
@@ -767,6 +775,36 @@ function overrideJsonProperties(originalObject, overrideObject) {
...
+  keyArray = Object.keys(originalObject);
+  for (let i = 0; i < keyArray.length; i++) {
+    key = keyArray[i];
+    property = originalObject[key];
+    overrideProperty = htGet(overrideTable,key);
+    if (!overrideProperty) {
+      overrideObject[key] = property;
+    } else {
+      if (Array.isArray(overrideProperty)) {
+        if (Array.isArray(property)) {
```
1000TurquoisePogs: If both ARENT arrays, you need an else... which does `overrideObject[key] = property;`
Because right now, you'd get nothing.

plugins/config/lib/configService.js
1000TurquoisePogs: Also, maybe print a debug-level warning if one is an array and one isnt... because something weird happened.

plugins/config/lib/configService.js
```diff
@@ -2264,6 +2337,21 @@ function getPluginConfiguration(identifier, pluginLocation, serverSettings,produ
 };
 exports.getPluginConfiguration = getPluginConfiguration;
 
+function getAllowedPlugins(scope, options, username) {
```
1000TurquoisePogs: You dont need SCOPE as a variable here because, if username == null|undefined, then SCOPE=INSTANCE, else SCOPE=USER

plugins/config/lib/configService.js
```diff
@@ -2264,6 +2337,21 @@ function getPluginConfiguration(identifier, pluginLocation, serverSettings,produ
...
+  let policy = AGGREGATION_POLICY_MERGE
+  let identifier = "org.zowe.zlux.bootstrap"
+  let filename = "allowedPlugins.json"
+  let relativePath = identifier + "/plugins"
+  let serverSettings = options.serverConfig
+  let productCode = "ZLUX"
+  let pluginLocation = "C:\\Users\\skeerthy\\post90\\zlux\\zlux-app-manager\\bootstrap"
```
1000TurquoisePogs: It's hard to get the plugin location from here, so I suggest moving this entire function out to webapp.js
Here's how it can work:
webapp has: id='org.zowe.zlux.bootstrap', filename='allowedPlugins.json', serverConfig.... but most importantly the location of org.zowe.zlux.bootstrap is found at https://github.com/zowe/zlux-server-framework/blob/12488f38c01ac051ee052dcf27101fd63aa8b25e/lib/webapp.js#L221
`plugins: function(plugins, options){`
The "plugins" argument is an array that contains objects which have `location` as the full path to the directory they are in.
`pluginDefs` doesnt, because it is intentionally stripped.

lib/webauth.js
sakeerthy: adding the res.status(401) breaks the login page since its querying it without authentication 

removing the else breaks logging in to a user with allwoed plugins, the allowed plugins doesn't show

plugins/config/lib/configService.js
```diff
@@ -2264,6 +2358,18 @@ function getPluginConfiguration(identifier, pluginLocation, serverSettings,produ
...
 exports.getPluginConfiguration = getPluginConfiguration;
 
+function getAllowedPlugins(options, username, identifier, pluginLocation) {
+  let filename = "allowedPlugins.json"
+  let relativePath = identifier + "/plugins"
+  let serverSettings = options.serverConfig
+  let productCode = options.pluginLoader.options.productCode
+  let directories = makeConfigurationDirectoriesStructInner(serverSettings, productCode, username, pluginLocation);
+  directories._pluginID = encodeDirectoryName(identifier)
+  return getMergeJson(relativePath, filename, directories)
```
1000TurquoisePogs: Scope is required but omitted

lib/webapp.js
```diff
@@ -233,21 +233,55 @@ const staticHandlers = {
...
+          if (plugin.pluginType === "bootstrap") {
+            pluginId = plugin.identifier
+            pluginLocation = plugins.location
+            break;
+          }
+        }
+        if (req.username) {
+          allowedPlugins = configService.getAllowedPlugins(options, req.username, pluginId, pluginLocation);
+        } else { 
+          allowedPlugins = configService.getAllowedPlugins(options, undefined, pluginId, pluginLocation);
```
1000TurquoisePogs: You don't need an if-else because if !req.username, req.username = undefined|null

plugins/config/lib/configService.js
```diff
@@ -1049,9 +1091,27 @@ function respondWithConfigFile(response, filename, resource, directories, scope,
         else {
           respondWithJsonError(response,"Resource not yet defined",HTTP_STATUS_NO_CONTENT,location);
         }
-      }); 
+      }, overrideJsonProperties); 
     }
     break;
+  case AGGREGATION_POLICY_MERGE:
+    {
+      getOverrideJsonAsync(lastPath,filename,directories,scope,(result)=> {
```
1000TurquoisePogs: This would be wrong - getOverrideJsonAsync doesn't do merge, it does override.

plugins/config/lib/configService.js
sakeerthy: https://github.com/zowe/zlux-server-framework/pull/153#discussion_r332650846

plugins/config/lib/configService.js
sakeerthy: i added the call back function to that function based on if it was merge or override

1000TurquoisePogs:
This PR removes the test scripts from running, which should be put back in shortly.

PR #154, opened 2019-10-18
CLI -D arguments

1000TurquoisePogs:
Allow overriding of any config attribute via java/C-like -D argument syntax.
TODOs and comments within about what this doesn't cover: not trying to make this completely faithful to JSON at this point, but just to cover strings well.
Check on: correct behavior for numbers & booleans
Needed by: https://github.com/zowe/zlux-app-server/pull/80/


utils/argumentParser.js
```diff
@@ -0,0 +1,193 @@
...
+  
+
+  Other limitations:
+   - Treats strings that start with [, end with ] to be arrays
+   - Arrays cannot have strings with ',' within them
+
+  TODO: support array insertion
+*/
+function resolveJson(argumentsObj, matchObj) {
+  const matchParts = matchObj.jsonName.split('.');
```
1000TurquoisePogs: Support for number & boolean???

PR #157, opened 2019-11-21
Do not merge override object if null due to bad parse

1000TurquoisePogs:


1000TurquoisePogs:
The case in which I needed this fix was if I had a JSON file that was invalid... the try-catch would return null on purpose, but the receiving end wouldn't check for null. Now it does, and safely moves on.

PR #160, opened 2019-12-20
Server Up message added

sakeerthy:

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -214,6 +214,17 @@ Server.prototype = {
     this.webServer.startListening(this.webApp.expressApp);
     let pluginsLoaded = 0;
     let pluginCount = 0;
+    let runningInstances = [];
+    if (webAppOptions.serverConfig.node.https) {
+      for (let ip of webAppOptions.serverConfig.node.https.ipAddresses) {
+        runningInstances.push("https://" + ip + ':' + httpsPort)
+      }
+    }
+    if (webAppOptions.serverConfig.node.http) {
+      for (let ip of webAppOptions.serverConfig.node.http.ipAddresses) {
+        runningInstances.push("http://" + ip + ':' + httpsPort)
+      }
+    }
     this.pluginLoader.on('pluginAdded', util.asyncEventListener(event => {
       pluginCount++;
       return this.pluginLoaded(event.data).then(() => {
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -255,6 +258,10 @@ Server.prototype = {
       });
       yield this.apiml.registerMainServerInstance();
     }
+    bootstrapLogger.log(bootstrapLogger.INFO,
```
1000TurquoisePogs: This is hit asynchronously to the events above, so it cannot be correct that the server is actually ready at this time.

lib/index.js
```diff
@@ -212,10 +212,13 @@ Server.prototype = {
...
     this.webApp = makeWebApp(webAppOptions);
     this.webServer.startListening(this.webApp.expressApp);
-    let pluginsLoaded = [];
+    let pluginsLoaded = 0;
+    let pluginCount = 0;
     this.pluginLoader.on('pluginAdded', util.asyncEventListener(event => {
+      pluginCount++;
       return this.pluginLoaded(event.data).then(() => {
         installLogger.info('Installed plugin: ' + event.data.identifier);
+        pluginsLoaded++;
```
1000TurquoisePogs: By looking at my logs, it appears this is a pretty late place in the plugin loading. If you were able to determine you loaded all plugins you wanted to load at this time, then you could print a final message.
Perhaps event.data could contain info that would tell you the % complete to be able to make that check.

lib/index.js
```diff
@@ -212,14 +212,31 @@ Server.prototype = {
...
       }, err => {
         installLogger.warn(`Exception occurred, plugin (${event.data.identifier}) installation skipped. `
-                           +`Message: ${err.message}`);
+                           +`Message: ${err.message}`
+                           + " Successful: (" + pluginsLoaded + "/" + event.count + ")"
+                           + " Attempted: (" + pluginCount + "/" + event.count + ")");
         installLogger.debug(err.stack);
+        if (pluginCount == event.count) {
+          installLogger.info(`Server is ready at ` +
+            `${webAppOptions.serverConfig.node.https.ipAddresses}:${webAppOptions.httpsPort || webAppOptions.httpPort}` +
```
1000TurquoisePogs: This uses https ipAddresses even if https is not used, but uses either https or http port.
To be concise, it would be best to prefer https but if not found, mention http.

1000TurquoisePogs: This also has weird output when there are multiple IP addresses.

1000TurquoisePogs:
Tested, works!

PR #162, opened 2020-01-01
Adding Internal-only services feature

1000TurquoisePogs:
Adding an ability for a service to be declared as only for use by other services internally.

This allows for a service to be callable by other services of that plugin via the "callService" function, but does not allow these services to be called from outside of the server. Attempting to call them from outside of the server will be met with a 403 return and an error message.
Since a service is a router and routers are attached internally & externally the same way, the way the server discriminates what is internal vs external in this case is that it checks for the request IP, as well as a code, intended to be only known to the server. If the IP is determined to come from the same host as the server, and the code is valid, then the request is passed on to the router, otherwise the 403 results.

How to make use of this:

Dataservices can now contain the attribute "internalOnly"=true/false. If false, current behavior is seen. If missing, treated as false. But if true, the above-described behavior occurs.

### DIFF
--- a/lib/clusterManager.js
+++ b/lib/clusterManager.js
@@ -82,14 +82,19 @@ if (cluster.isMaster) {
       if (crypto_err) {
         return errorHandler(crypto_err);
       }
-      this.secret = buffer.toString('hex');
-      //console.log('secret: ' + this.secret);
+      const secret = buffer.toString('hex');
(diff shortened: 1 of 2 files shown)


1000TurquoisePogs:
This has been reviewed by people who originally requested it for their development, and tested by my team, so I'm merging it in.

PR #169, opened 2020-01-28
Combine apiml-auth and zss-auth into one

1000TurquoisePogs:
Making a single plugin, called "sso-auth", which is of type "saf".
This is a plugin which will be for SSO in the long run, but it also is able to do just apiml or just zss if one is missing. If both are present, it will use apiml token to give to zss for SSO.
This should not be used as the default auth yet until the standard configuration of Zowe is set to have zss use the SSO token. But, at some point we should switch over to this plugin and stop including zss-auth and apiml-auth.

The code is nearly copy-pasted, in which I put apiml-auth into "apimlHandler.js", and zss-auth into "zssHandler.js", and conditionally call their functions in "ssoAuth.js"

Potential problem: this plugin is of security category "saf", not "zss" or "apiml", so if plugins explicitly requested types "zss" or "apiml" then they would fail. It may be necessary to enhance our api to have a security plugin be declared as multiple types, such as `"authenticationCategory": ["saf","zss","apiml"]` but this is somewhat misleading at runtime if zss is not present, so perhaps even a constructor return statement is needed.

How I tested this:
1. Ran zss without apiml present... I can login and do everything that doesnt rely on apiml endpoints
2. Ran apiml without zss present... I can login and do everything that doesnt rely on zss apis

### DIFF
--- a/plugins/sso-auth/lib/apimlHandler.js
+++ b/plugins/sso-auth/lib/apimlHandler.js
@@ -1,0 +1,201 @@
+/*
+  This program and the accompanying materials are
+  made available under the terms of the Eclipse Public License v2.0 which accompanies
+  this distribution, and is available at https://www.eclipse.org/legal/epl-v20.html
+  
+  SPDX-License-Identifier: EPL-2.0
+  
+  Copyright Contributors to the Zowe Project.
+*/
+
+const Promise = require('bluebird');
+const https = require('https');
+const fs = require('fs');
+
+/*495 minutes default session length for zosmf
+ * TODO: This is the session length of a zosmf session according to their documentation.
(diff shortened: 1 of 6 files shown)


plugins/sso-auth/lib/apimlHandler.js
```diff
@@ -0,0 +1,203 @@
...
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'Content-Length': data.length
+        },
+        agent: this.httpsAgent
+      }
+
+      const req = https.request(options, (res) => {
+        res.on('data', (d) => {});
```
timgerstel: Is this needed if nothing is done with the parameter returned?

PR #170, opened 2020-01-30
Get correct count of successfully loaded plugins

1000TurquoisePogs:
The messages for plugin loading status did not account for plugins that existed but could not be loaded due to a failure early on.
This fixes one such problem: if a plugin failed to get dataservices set up, it would not be in the array of counted plugins. Now it will be, but will contain an error object, so it can be counted as an error and excluded from further processing.

As an example, before I was reading 15/15 plugins successfully loaded, now I read 15/17 plugins successfully loaded, because one had a require() exception and another was pointing to a folder that didnt exist.

New messages read as:
```
2020-01-30 23:22:44.559 <ZWED:29881> sgrady INFO (_zsf.install,index.js:246) Plugin (org.zowe.terminal.tn3270) loaded.  Successful: 81% (13/16) Attempted: 81% (13/16)
2020-01-30 23:22:44.559 <ZWED:29881> sgrady WARN (_zsf.install,index.js:232) Plugin (org.zowe.error) loading failed. Message: "Cannot find module '/sgrady/error/lib/starter.js'" Successful: 81% (13/16) Attempted: 88% (14/16)
```

### DIFF
--- a/lib/plugin-loader.js
+++ b/lib/plugin-loader.js
@@ -580,23 +580,27 @@ PluginLoader.prototype = {
       pluginBasePath = this.options.relativePathResolver(pluginBasePath, process.cwd());
     }
     if (!fs.existsSync(pluginBasePath)) {
-      throw new Error(`${pluginDescriptorFilename}: No plugin directory found at`
-        + ` ${pluginPtrDef.pluginLocation}`);
+      return {location: pluginBasePath,
+              identifier: pluginPtrDef.identifier,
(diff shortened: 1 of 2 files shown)


lib/plugin-loader.js
```diff
@@ -682,14 +692,13 @@ PluginLoader.prototype = {
         }
       } catch (e) {
         console.log(e);
```
DivergentEuropeans: Is there a reason why we are outputting the error to a console log?

PR #172, opened 2020-02-11
Add plugin logout ability

1000TurquoisePogs:
The auth-logout endpoint never reached the security plugins, in case there was something else they should be doing on the backend during the logout process.
Now, if the plugins support it, they are called. This is similar to the logic used for login, except that regardless of what the plugins do, the server core still deletes the session storage area at the end.

### DIFF
--- a/plugins/sso-auth/lib/apimlHandler.js
+++ b/plugins/sso-auth/lib/apimlHandler.js
@@ -68,6 +68,64 @@ class ApimlHandler {
     }
   }
 
+  logout(request, sessionState) {
+    return new Promise((resolve, reject) => {
+      const gatewayUrl = this.gatewayUrl;
+      const options = {
+        hostname: this.apimlConf.hostname,
+        port: this.apimlConf.gatewayPort,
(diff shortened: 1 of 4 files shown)


plugins/sso-auth/lib/apimlHandler.js
```diff
@@ -68,6 +68,64 @@ class ApimlHandler {
...
+        path: '/api/v1/apicatalog/auth/logout',
+        method: 'POST',
+        headers: {
+          'apimlAuthenticationToken': sessionState.apimlToken
+        },
+        agent: this.httpsAgent
+      }
+
+      const req = https.request(options, (res) => {
+        res.on('data', (d) => {});
```
timgerstel: Is this statement needed if it has no functionality?

1000TurquoisePogs: Not necessarily true... it may need an event listener to function properly.
I just copy-pasted from the previously working code, so I'm not going to mess with it right now.

PR #178, opened 2020-02-26
Feature/add server side storage

DivergentEuropeans:
- Adding a server side storage API for cluster support with set, get, delete
- If the storage hasn't loaded yet, the functionality is still available and any changes get updated retroactively once storage has finished loading (so we don't slow down app startup by having them wait for cluster to make the storage)

image

Sample Angular App PR: https://github.com/zowe/sample-angular-app/pull/32/files


lib/webapp.js
```diff
@@ -134,6 +149,51 @@ DataserviceContext.prototype = {
...
+      process.clusterManager.setStorageValue(id, value);
+    }
+
+    clusterStorage.setStorage = function (dict) {
+      clusterStorage.storageDict = dict;
+      process.clusterManager.setStorage(dict);
+    }
+
+    clusterStorage.addStorage = function (dict) {
+      clusterStorage.storageDict = { ...clusterStorage.storageDict, ...dict };
```
timgerstel: Have you tested how this merges objects with depth n > 2,
aka would:

```
let obj1 = {
     key1: 'value1',
     key2: {
       subkey1: 'value2'
     }
}
```

merge properly with

```
let obj2 = {
     key1: 'differentValue',
     key2: {
          subkey2: {
                subkey3: 'value3'
           }
     }
}
```

lib/clusterManager.js
```diff
@@ -521,6 +583,88 @@ if (cluster.isMaster) {
...
+    );
+  }
+
+  ClusterManager.prototype.getStorageValue = function(id) {
+    return this.callClusterMethodRemote(null, "clusterManager", "getStorageValue", [id],
+    function(resultHandler) { 
+      return resultHandler[0]; // Return the first arg (storage object) of getStorage
+    },
+    function(e) {
+      clusterLogger.warn("Unable to retrieve storage object from cluster", e);
```
timgerstel: Should be 'unable to retrieve get storage value'

lib/webapp.js
DivergentEuropeans: @timgerstel That's a good point, and if I were to guess I would say it wouldn't handle this because I remember @1000TurquoisePogs having a problem doing this and not being able to find a JS way (and had to import some package?)

lib/clusterManager.js
DivergentEuropeans: Thanks, good catch.

lib/clusterManager.js
```diff
@@ -187,7 +192,63 @@ if (cluster.isMaster) {
     this.overrideFileConfig = override;
   }
 
-   ClusterManager.prototype.notifyWorkersForAddingPlugin = function(resultHandler, indexInCluster) {
+  /* Having gotten here, we are already at the cluster, but it's safer to validate anyway
+    i. e. "if (this.storage) ..." since only the master cluster has storage */
+  ClusterManager.prototype.getStorage = function(resultHandler) {
+    if (this.storage) {
```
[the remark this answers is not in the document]
DivergentEuropeans: I think I'll leave it the way I have it now for the sake of code clarity (and consistency, since we have this simple to read way nearly everywhere so we'd want to go back and change all those occurrences as well)

lib/clusterManager.js
```diff
@@ -187,7 +192,63 @@ if (cluster.isMaster) {
...
+      return resultHandler(zluxUtil.DataserviceStorage());
+    }
+  }
+
+  ClusterManager.prototype.setStorageValue = function(id, value, resultHandler) {
+    if (this.storage) {
+      this.storage.setStorageValue(id, value);
+      return resultHandler(true);
+    } else {
+      console.error("Master cluster not found to save storage for id: " + id + ", value: " + value);
```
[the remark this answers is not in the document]
DivergentEuropeans: Yes, I will thanks @timgerstel

lib/webapp.js
DivergentEuropeans: I figured it out. Will add my own method to do a recursive merge

lib/clusterManager.js
```diff
@@ -557,6 +702,35 @@ ClusterManager.prototype.notifyOthers = function (notifyName, args, callback, on
   }
 }
 
+/* Performs a recursive merge on objects, which is different from spread/Object.assign operations 
+that do not merge n > 2 levels down. Does not modify original objects. */
+function recursiveMerge(...objects) {
```
timgerstel: Have you tested what happens with objects that have the same key but different value?

Example:

```
let obj1 = {
  key1: {
    key2: 'value',
    key3: 'value'
  }
}
```

```
let obj2  = {
  key1: {
    key2: 'value',
    key4: 'value'
  }
}
```


Will obj 1 no longer have key3 if merged?  Hopefully you tested this

DivergentEuropeans: Yes, I did, it's in the Sample App PR

lib/webapp.js
```diff
@@ -75,6 +75,21 @@ function DataserviceContext(serviceDefinition, serviceConfiguration,
...
   this.serviceConfiguration = serviceConfiguration;
   this.plugin = pluginContext;
+
+  /* By default, we init storage object to a default one as to not slow down app init by
+  waiting for the real one (or if we are not in cluster mode) */
+  this.storage = zluxUtil.DataserviceStorage();
+
+  if (process.clusterManager) { // Cluster mode
+    createDataserviceStorage().then((data) => {
+      /* Then, once the cluster is done creating its own storage, we merge the local one
```
1000TurquoisePogs: The idea of needing to merge local & cluster storage can be entirely avoided if there is code guarantees that the cluster finishes its creation prior to this section running. It probably does, but a solution to ensure it would simplify.

lib/webapp.js
1000TurquoisePogs: I'm just not sure why this function exists, though.

lib/webapp.js
```diff
@@ -134,6 +149,51 @@ DataserviceContext.prototype = {
...
+
+    clusterStorage.deleteStorageValue = function (id) {
+      delete clusterStorage.storageDict[id];
+      process.clusterManager.deleteStorageValue(id);
+    }
+
+    contentLogger.debug("App context is loaded with storage object: ", clusterStorage);
+    return clusterStorage;
+  })
+  .catch( function(e) { // Chances are, if storage failed to load, then the logger may have failed to load
```
1000TurquoisePogs: If you fail to get the cluster storage, then plugins that depend on this storage feature will not work as expected. It is time to stop the server.

lib/webapp.js
```diff
@@ -134,6 +149,51 @@ DataserviceContext.prototype = {
...
+      process.clusterManager.setStorage(dict);
+    }
+
+    clusterStorage.addStorage = function (dict) {
+      clusterStorage.storageDict = { ...clusterStorage.storageDict, ...dict };
+      process.clusterManager.addStorage(dict);
+    }
+
+    clusterStorage.getStorageValue = function (id) {
+      process.clusterManager.getStorageValue(id).then((value) => {
```
1000TurquoisePogs: This is what I was looking for, cool.

lib/webapp.js
```diff
@@ -134,6 +149,51 @@ DataserviceContext.prototype = {
...
+        return value;
+      });
+    }
+
+    clusterStorage.deleteStorageValue = function (id) {
+      delete clusterStorage.storageDict[id];
```
1000TurquoisePogs: This seems wrong to me, because it assumes there are 2 storage objects at this time. There should be 1: on the cluster manager

lib/webapp.js
```diff
@@ -134,6 +149,51 @@ DataserviceContext.prototype = {
...
+    
+    clusterStorage.getStorage = function () {
+      return process.clusterManager.getStorage().then((storage) => {
+        clusterStorage.storageDict = storage.storageDict;
+        return storage.storageDict;
+      });
+    }
+
+    clusterStorage.setStorageValue = function (id, value) {
+      clusterStorage.storageDict[id] = value;
```
1000TurquoisePogs: This doesn't make sense: it looks like there are 2 objects when there should be 1: the clustermanager's store. what is clusterStorage.storageDict doing?

lib/webapp.js
```diff
@@ -75,6 +75,21 @@ function DataserviceContext(serviceDefinition, serviceConfiguration,
   this.serviceDefinition = serviceDefinition;
   this.serviceConfiguration = serviceConfiguration;
   this.plugin = pluginContext;
+
+  /* By default, we init storage object to a default one as to not slow down app init by
+  waiting for the real one (or if we are not in cluster mode) */
+  this.storage = zluxUtil.DataserviceStorage();
```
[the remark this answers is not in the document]
DivergentEuropeans: I was under the impression that, that was the whole point. To have storage accessible to any plugin, at the cluster level, but we can make it more rigid as to not have the plugins conflict with each other

lib/webapp.js
DivergentEuropeans: The cluster does not finish its creation prior to this section running and it would never have the storage that could have been added while the cluster didn't finish loading yet. That was the point of merging them together because at this point, there is
1. the master storage in cluster has loaded
2. the master storage in cluster probably does not have any of the storage updates of the local storage, if the local storage was messed with prior to the cluster finishing loading.

This was all set up to NOT wait for the cluster, since that would slow down app init, am I misunderstanding? Unless we do want to make apps load slower by having wait for the cluster @1000TurquoisePogs

DivergentEuropeans:
Hey, thanks for the large review. I would highly recommend checking out my Sample Angular App PR (more specifically here: https://github.com/zowe/sample-angular-app/pull/32/files#diff-06f7c90458492ea6721326ccfa10e860) which I think will clarify many of your questions. That being said, looking at it, these method names are confusing and I may need to think of better names for them.

> I may need this explained to me in case there is novelty or subtlety that I am missing.
> This PR appears to have solved the communication part of cluster storage, but requires cleanup to become usable by plugins.

Check out the Sample Angular App PR, and let me know if it makes sense to you and if the usage is valid

> A storage API at minimum needs get and set. You added delete which is even better.
> But here there is getStorageValue, getStorage, addStorage

getStorageValue get's the value of one specific object. So for example if all storage consists of: 
{ hello: okay, hello2: notokay }

Then getStorageValue(hello) would return 'okay' instead of all storage

getStorage gets all the storage (the whole storage object)

addStorage is used to add to the storage by supplying another storage object, but it's not a set so it doesn't remove the existing stuff. It just merges them together.

> It does not make sense to me when I look at the details, so I assume it will make less sense if people only look at the function argument & names alone.
> This is where code comments could clear things up but let me guess what's going on:
> 

Definitely need to add more!

> There is some initialization concern that you want to solve by combining temporary storage with permanent storage once it becomes available.
> 

Yes

> In the cluster, the model that we need is that the cluster holds the most current information, and therefore the workers should not store information at all except for maybe as a startup buffer awaiting the cluster to be ready.
> The reason why the workers should not store data is because it duplicates it. If the cluster crashes, the entire server crashes. If the worker crashes, the cluster must start up a new worker with the state the old one has. So, the more you store at the worker level the more you have to update, which is hard on the CPU.
> 

Yeah I fundamentally misunderstood what you wanted here @1000TurquoisePogs Storage exists in two places, 1. the app context (webapp) and 2. the master cluster
In our conversations, I thought that I told you I was copying them over in two places, with the cluster being the "master" that is the real storage, and app context will have its own copy available for plugins to use. As plugins use them, the cluster one then gets updated after, but the app context is the one that stores data, verifying with the cluster.

In retrospect, this is a pretty heavy way of doing things and I will make it better by referencing 1 object in 2 different places instead of referencing 2 objects in 2 different places

> So, getStorage and addStorage, if they are just for cluster-worker syncing, they should not be functions because they should only be used once: at initialization, if required. Post-init, in cluster mode the storage object should be a pass-through to-from the cluster. Without cluster mode, it's a local storage object of the same shape: get, set, and delete. Therefore also you do not need the distinction between getStorageValue and getStorage. It would just be storage.get
> 

Right, that makes sense. However, I disagree with the last sentence @1000TurquoisePogs What if we want to, at a plugin level, update only 1 setting/storage setting as opposed to updating the whole thing. This means that if we wish to add storage, we would have to KNOW the whole thing, then add to it, and then set the whole thing, instead of just setting one key. Using 1 whole object sounds very heavy?

> Finally, the storage are must do its best to isolate between plugins.

Will do

> The best way to do this is to give the plugin a wrapper to the storage area which is basically a key to the pluginID, so that the plugin only sees its own storage, not those of the others.

lib/webapp.js
DivergentEuropeans: Yep, there are two. My misunderstanding! See my reply.

lib/webapp.js
DivergentEuropeans: Yep, there are two. My misunderstanding! See my reply.

lib/webapp.js
DivergentEuropeans: Plugins will still have storage, they just won't have the cluster storage. Still fatal?

lib/clusterManager.js
```diff
@@ -189,7 +194,64 @@ if (cluster.isMaster) {
...
+
+  ClusterManager.prototype.getStorageValue = function(id, resultHandler) {
+    if (this.storage) {
+      return resultHandler(this.storage.getObject(id));
+    } else {
+      return resultHandler(null);
+    }
+  }
+
+  ClusterManager.prototype.deleteStorageValue = function(id) {
```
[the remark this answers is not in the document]
DivergentEuropeans: 1. Good catch! @1000TurquoisePogs Is "deleteStorageValue"
2. Solely because I forgot to add a callback. I will make it async, so it has a callback, just as the others do, as we will need our users to be informed when these cluster operations are finished

lib/clusterManager.js
```diff
@@ -189,7 +194,64 @@ if (cluster.isMaster) {
...
+      return resultHandler(zluxUtil.DataserviceStorage());
+    }
+  }
+
+  ClusterManager.prototype.setStorageValue = function(id, value, resultHandler) {
+    if (this.storage) {
+      this.storage.setStorageValue(id, value);
+      return resultHandler(true);
+    } else {
+      clusterLogger.warn("ZWED0160W - Master cluster not found to save storage for id: " + id + ", value: " + value);
```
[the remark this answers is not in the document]
DivergentEuropeans: The cluster will have storage because it is made upon startup. The plugins will always have storage because even before cluster storage is checked for, a default local one is made, so no plugin functionality should break. If we create the storage after encountering an error, then it may be because this method was called at startup, so setStorage for an object that doesn't exist yet will turn up false and creating a storage object here would be creating a second object and thus mess with things terribly

lib/webapp.js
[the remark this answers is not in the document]
DivergentEuropeans: No problem, I thought we wanted indiscriminate storage originally in the criteria. I'll change it to organize itself at the plugin level

lib/webapp.js
```diff
@@ -134,6 +149,51 @@ DataserviceContext.prototype = {
...
+        return storage.storageDict;
+      });
+    }
+
+    clusterStorage.setStorageValue = function (id, value) {
+      clusterStorage.storageDict[id] = value;
+      process.clusterManager.setStorageValue(id, value);
+    }
+
+    clusterStorage.setStorage = function (dict) {
```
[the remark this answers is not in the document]
DivergentEuropeans: Changed, thank you

lib/clusterManager.js
```diff
@@ -189,7 +195,82 @@ if (cluster.isMaster) {
...
+  ClusterManager.prototype.getStorageAll = function(id, resultHandler) {
+    if (this.storage) {
+      return resultHandler(this.storage.getStorageAll(id));
+    } else {
+      return resultHandler(zluxUtil.DataserviceStorage(id).getStorageAll());
+    }
+  }
+
+  ClusterManager.prototype.getStorageCluster = function(resultHandler) {
+    if (this.storage) {
```
1000TurquoisePogs: All of these if-elses  should not happen. If you do not have this.storage at this time, you should stop the server with a message saying what went wrong (storage didnt exist) and how to fix it (increase timeout in server config file)

lib/clusterManager.js
```diff
@@ -523,6 +604,93 @@ if (cluster.isMaster) {
...
+
+  ClusterManager.prototype.getStorageAll = function(id) {
+    return this.callClusterMethodRemote(null, "clusterManager", "getStorageAll", [id],
+      function(resultHandler) {
+        return resultHandler[0]; // Return the first arg (storage object) of getStorageAll
+      },
+      function(e) {
+        clusterLogger.warn("ZWED0164W - Unable to retrieve storage object from cluster", e);
+        return e;
+      }, 5000 /* We need to give ample timeout time since this method is used to verify
```
1000TurquoisePogs: This needs to be configurable by server.json, something in the "node" object, but maybe this needs to be in a section specific to clustering, such as node.cluster.storageTimout=5000

lib/webapp.js
```diff
@@ -134,6 +151,41 @@ DataserviceContext.prototype = {
   makeErrorObject: zluxUtil.makeErrorObject
 };
 
+function createDataserviceStorage(pluginId) {
+  
+  return process.clusterManager.getStorageAll(pluginId).then(function (clusterStorage) {
+
+    /* Get the whole plugin storage */
+    clusterStorage.getStorageAll = function () {
```
1000TurquoisePogs: Dont modify clusterStorage here, just create a brand new object with these functions and return it... basically the object returned by getStorageAll here is pointless to you here.

lib/webapp.js
```diff
@@ -134,6 +151,41 @@ DataserviceContext.prototype = {
...
+    /* Set the whole plugin storage object */
+    clusterStorage.setStorageAll = function (dict) {
+      return process.clusterManager.setStorageAll(pluginId, dict);
+    }
+
+    clusterStorage.setStorageByKey = function (key, value) {
+      return process.clusterManager.setStorageByKey(pluginId, key, value);
+    }
+
+    clusterStorage.deleteStorageByKey = function (key) {
```
1000TurquoisePogs: Change these all to:
get
getAll
set
setAll
delete
(deleteAll???)

lib/webapp.js
```diff
@@ -134,6 +151,41 @@ DataserviceContext.prototype = {
...
+      return process.clusterManager.deleteStorageByKey(pluginId, key);
+    }
+
+    contentLogger.debug("'" + pluginId + "' context is loaded with storage object: ", clusterStorage);
+    return clusterStorage;
+  })
+  .catch(function(e) { // Chances are, if storage failed to load, then the logger may have failed to load
```
1000TurquoisePogs: If you do somehow catch here, something bad has happened so the server should stop instead by sending a specific error number up to the cluster.

lib/assets/i18n/log/messages_en.json
```diff
@@ -276,6 +276,16 @@
...
+  "ZWED0160W":"RESERVED: Master cluster not found to save storage for id: %s, value: %s",
+  "ZWED0161W":"RESERVED: Master cluster not found to save storage",
+  "ZWED0162W":"RESERVED: Master cluster not found to add storage",
+  "ZWED0163W":"RESERVED: Master cluster was not found to delete storage for id: %s",
+  "ZWED0164W":"RESERVED: Unable to retrieve storage object from cluster %s",
+  "ZWED0165W":"RESERVED: Unable to retrieve storage data from cluster promise %s",
+  "ZWED0166W":"RESERVED: Error updating the storage: %s",
+  "ZWED0167W":"RESERVED: Error adding to the storage: %s",
+  "ZWED0168W":"RESERVED: Unable to retrieve storage value from cluster %s",
+  "ZWED0169W":"RESERVED: Error deleting the storage with id: %s %s",
```
1000TurquoisePogs: Need a new message here about timeout of storage if encountered...

1000TurquoisePogs:
After cleanup & retest, this is merge-ready.

PR #181, opened 2020-03-11
Password Reset

sakeerthy:
WIP

Goes along with:
https://github.com/zowe/zss/pull/161
https://github.com/zowe/zowe-common-c/pull/128
https://github.com/zowe/zlux-app-server/pull/106
https://github.com/zowe/zss-auth/pull/29
https://github.com/zowe/zlux-platform/pull/52
https://github.com/zowe/zlux-app-manager/pull/193


1000TurquoisePogs:
I think this is the right approach. Just awaiting ZSS stuff.

lib/webapp.js
```diff
@@ -1303,6 +1303,8 @@ WebApp.prototype = {
     
     this._installRootService('/auth', 'post', this.auth.doLogin, 
         {needJson: true, needAuth: false, isPseudoSso: true});
+    this._installRootService('/auth-password', 'post', this.auth.doPasswordReset,
+        {needJson: true, needAuth: false, isPseudoSso: true});
```
1000TurquoisePogs: `isPseudoSso` may cause a problem. More below.

lib/webauth.js
```diff
@@ -207,6 +207,33 @@ module.exports = function(authManager, cookiePort, isSecurePort) {
     return;
   });
 
+  const _passwordResetInner = Promise.coroutine(function*(req, res, type) {
+   const handlers = getRelevantHandlers(authManager, req.body);
+   const result = new LoginResult();
+   for (const handler of handlers) {
```
1000TurquoisePogs: If you do this, it means that you are requesting a password reset from every handler.
The way this handler stuff is supposed to work is that maybe you have a handler for zos, facebook, and windows logins. You may want to login to all 3 at once. You do not want to password reset all 3 at once.
I think you should require that req.body defines exactly 1 handler, or that the URL contains a handler ID, or something like that.
Otherwise the user experience will be very strange.

Practically speaking, what will happen on 1.9 is you have 2 handlers: zss-auth and apiml-auth
zss-auth will do what you want.
apiml-auth will have canResetPassword == false, which will return success:false, which will make the UI show an error.
So, password reset will not work in that default setup. You need to limit it to zss for now.

lib/webauth.js
```diff
@@ -277,7 +302,12 @@ module.exports = function(authManager, cookiePort, isSecurePort) {
         initZLUXSession(req);
         req.session.zlux.expirationTime = Date.now() + result.expms;
       }
-      
+      if (result.categories.zss) {
```
1000TurquoisePogs: dont hardcode this.
Just make it so that if any plugin has reason "Expired password" that a 428 is set. You need to go through the list.
Also, you need 428 to be a constant.

lib/webauth.js
```diff
@@ -207,6 +210,36 @@ module.exports = function(authManager, cookiePort, isSecurePort) {
     return;
   });
 
+  const _passwordResetInner = Promise.coroutine(function*(req, res, type) {
+    const handlers = getRelevantHandlers(authManager, req.body);
```
1000TurquoisePogs: More correct to just use authManager.getAllHandlers() here since it is deliberate rather than accidental, since the behavior you observe is I think accidental currently.

lib/webauth.js
```diff
@@ -277,6 +308,16 @@ module.exports = function(authManager, cookiePort, isSecurePort) {
...
+        if (authObject.success == false) {
+          for (let [authPlugin, authPluginObject] of Object.entries(authObject.plugins)) {
+            if (authPluginObject.reason == EXPIRED_PASSWORD) {
+              res.status(HTTP_STATUS_PRECONDITION_REQUIRED).json(result);
+              return;
+            }
+          }
+        }
+      }
       result.defaultCategory = authManager.defaultType;
```
1000TurquoisePogs: Put this above your for loop, because it should be in every reponse from now on.

PR #197, opened 2020-04-17
Object.values did not exist in node6

1000TurquoisePogs:


DivergentEuropeans:
--harmony flag works for me for Node 6.5.0

1000TurquoisePogs:
This appears to be my mistake: I was changing around flags and may have forgotten --haromy, which lead me to believe we broke v6 support when we did not.

PR #199, opened 2020-04-22
PoC for Leader Election

lchudinov:
### New environment variables shared across zlux instances:  
* ZLUX_RAFT_CLUSTER_ENABLED, set to TRUE or YES
* ZOWE_INSTANCE, used in session cookie name
* ZOWE_SESSION_SECRET - temporary solution
* ZLUX_RAFT_CLUSTER_SIZE - number of peers in zlux app server cluster, default - 3
* ZLUX_RAFT_MAX_LOG_SIZE - max Raft log size, default is 100 items
* ZLUX_RAFT_MIN_ELECTION_TIMEOUT - min election timeout in milliseconds, default 1000 ms
* ZLUX_RAFT_MAX_ELECTION_TIMEOUT - max election timeout in milliseconds, default 2000 ms
* ZLUX_RAFT_HEARTBEAT_INTERVAL - heartbeat interval in milliseconds, default 200 ms

### How to test a failover
0. Set global environment variables
```
ZLUX_RAFT_CLUSTER_ENABLED=TRUE
ZOWE_INSTANCE="my_zowe"
ZOWE_SESSION_SECRET="my-secret"
ZLUX_RAFT_CLUSTER_SIZE=3
```
0.  Compile Sample Angular App using branch https://github.com/lchudinov/zowe-sample-angular-app/tree/feature/add-storage-test
1. Setup three zlux instances with Mediation layer enabled
2. Run API Mediation Layer
3. Run all three instances
4. Open Eureka UI at https://localhost:10011 (user/user) and wait until the instances register in Eureka, find each other, and perform Leader Election. 
5. When Leader Election is done you will see one ZLUX instance with status `UP` and the other two instances with status `OUT_OF_SERVICE`, e.g. 
image
6. Open Zlux UI at https://localhost:10010/ui/v1/zlux/ZLUX/plugins/org.zowe.zlux.bootstrap/web/, run Sample Angular App.
7. Add a new key-value pair into a cluster storage using Sample Angular App UI: 
storage-test
8.  Stop Leader instance
9. Open Eureka UI at https://localhost:10011 and check that a new Leader has been elected: 
image Note that old leader might appear in status `OUT_OF_SERVICE` for some time.
7. Ensure that Zlux UI still works. Press `F5`. UI must not show a login dialog because your session has been replicated on all servers.
8. Open Sample Angular App and check that cluster storage has been replicated:
storage-test-2


lib/apiml.js
```diff
@@ -202,6 +203,93 @@ ApimlConnector.prototype = {
...
   
+  getInstanceId() {
+    return this.zluxServerEurekaClient.config.instance.instanceId;
+  },
+  
+  getZluxInstances() {
+    const eurekaClient = this.zluxServerEurekaClient;
+    return new Promise((resolve, reject) => {
+      eurekaClient.once('registryUpdated', () => {
+        const zluxInstances = eurekaClient.getInstancesByVipAddress('ZLUX');
```
1000TurquoisePogs, line 214: i don't know about the case sensitivity. In some parts we specify ZLUX and it appears as zlux. This may be a eureka bug.

lib/apiml.js
```diff
@@ -202,6 +203,93 @@ ApimlConnector.prototype = {
     });
   },
   
+  getInstanceId() {
+    return this.zluxServerEurekaClient.config.instance.instanceId;
+  },
+  
+  getZluxInstances() {
+    const eurekaClient = this.zluxServerEurekaClient;
+    return new Promise((resolve, reject) => {
```
1000TurquoisePogs, line 212: because `registryFetchInterval=30000`, does this really wait up to 30 seconds to resolve? is that ok for how the code works? i will read more below.

lib/apiml.js
```diff
@@ -202,6 +203,93 @@ ApimlConnector.prototype = {
...
+        resolve(zluxInstances);
+      });
+    });
+  },
+  
+  waitUntilZluxClusterIsReady(clusterSize) {
+    return this.getZluxInstances().then(
+      zluxInstances => {
+        if (zluxInstances.length < clusterSize) {
+          return this.waitUntilZluxClusterIsReady(clusterSize);
```
1000TurquoisePogs, line 226: what does the word cluster mean here?
does it mean the app-server that is running in cluster mode and has a number of workers?
or does it mean a group of multiple app servers?

1000TurquoisePogs: if this is about the number of workers, i think the code must be changed to only need 1 eureka client per app server, by using the cluster master to handle it.

lib/apiml.js
```diff
@@ -202,6 +203,93 @@ ApimlConnector.prototype = {
...
+  
+  _overrideStatus(status) {
+    return this._overrideStatusForInstance(this.zluxServerEurekaClient.instanceId, status);
+  },
+  
+  _overrideStatusForInstance(instanceId, status) {
+    return new Promise((resolve, reject) => {
+      this.zluxServerEurekaClient.eurekaRequest({
+        method: 'PUT',
+        uri: `${this.zluxServerEurekaClient.config.instance.app}/${instanceId}/status?value=${status}`
```
1000TurquoisePogs, line 261: is OUT_OF_SERVICE a special name to eureka/apiml, or is it just something that was chosen for the app-server

lib/clusterManager.js
```diff
@@ -120,6 +120,10 @@ if (cluster.isMaster) {
     var inits = [];
     inits.push(this.initSecret.bind(this));
     inits.push(this.initSecret.bind(this));
+    if (process.env.ZOWE_SESSION_SECRET) {
```
1000TurquoisePogs, line 123: is this actually secure, or can it be spied on? at least, i think we need to change zlux-app-server which prints all env vars at startup. It should omit printing things that say password or secret.

lib/clusterManager.js
```diff
@@ -120,6 +120,10 @@ if (cluster.isMaster) {
...
+      this.secret = process.env.ZOWE_SESSION_SECRET;
+      inits.pop();
```
1000TurquoisePogs, line 125: that pops loopbackSecret, not secret.
the initSecret function is weird.
loopbackSecret is a way to make it so that some dataservices can only be reached internally instead of by a browser.

lib/util.js
```diff
@@ -38,7 +38,9 @@ const loggers = {
   network: global.COM_RS_COMMON_LOGGER.makeComponentLogger("_zsf.network"),
   langManager: global.COM_RS_COMMON_LOGGER.makeComponentLogger("_zsf.lang"),
   clusterLogger: global.COM_RS_COMMON_LOGGER.makeComponentLogger("_zsf.cluster"),
-  storeLogger: global.COM_RS_COMMON_LOGGER.makeComponentLogger("_zsf.store")
+  storeLogger: global.COM_RS_COMMON_LOGGER.makeComponentLogger("_zsf.store"),
+  syncLogger: global.COM_RS_COMMON_LOGGER.makeComponentLogger("_zsf.sync"),
+  raftLogger: global.COM_RS_COMMON_LOGGER.makeComponentLogger("_zsf.raft"),
```
1000TurquoisePogs, line 43: remember to document in the github.com/zowe/zlux/wiki page.

lib/webapp.js
```diff
@@ -671,6 +679,16 @@ const staticHandlers = {
     });
     return router;
   },
+  
+  raft(webApp) {
+    const router = express.Router();
+    webApp.expressWs.applyTo(router);
+    router.ws('/raft', (ws, req) => {
+      console.log(`client connected to raft service`);
```
1000TurquoisePogs, line 687: will need to convert this to use of the logger

lib/apiml.js
```diff
@@ -202,6 +203,93 @@ ApimlConnector.prototype = {
     });
   },
   
+  getInstanceId() {
+    return this.zluxServerEurekaClient.config.instance.instanceId;
```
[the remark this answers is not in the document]
lchudinov: It is `instanceId` for Eureka. It is on form of ```instanceId: `${this.hostName}:zlux:${port}` ```.

lib/apiml.js
lchudinov: We register zlux-app-server with `vipAddress` `zlux` in lowercase. However, Eureka coverts it to uppercase so you can find the zlux-app-server instances only by `ZLUX` in uppercase.

lib/apiml.js
lchudinov: The Eureka client will fetch Eureka registry every 30 seconds so we need to wait less than 30 seconds in 99.9% of cases.

lib/apiml.js
lchudinov: It is about a number of zlux-app-server instances. Yes, I should give it more apt name.

lib/apiml.js
lchudinov: Eureka server recognizes a few statuses — `UP`, `DOWN`, `STARTING`, `OUT_OF_SERVICE` and `UNKNOWN`.

lib/clusterManager.js
lchudinov: I created a PR https://github.com/zowe/zlux-app-server/pull/120

lib/clusterManager.js
lchudinov: I checked that it works. I need a single call to `initSecret()` and the call initializes `loopbackSecret` because `secret` is already initialized.

lib/util.js
lchudinov: Sure.

lib/webapp.js
lchudinov: This message must be removed.

lib/apiml.js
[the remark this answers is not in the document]
1000TurquoisePogs: ok, that's good.

lib/apiml.js
1000TurquoisePogs: It's best to define those as consts somewhere so that people will know what the choices are for future code.

lib/index.js
```diff
@@ -185,6 +189,9 @@ Server.prototype = {
     const httpPort = wsConfig.http ? wsConfig.http.port : undefined;
     const httpsPort = wsConfig.https ? wsConfig.https.port : undefined;
     const cookiePort = httpsPort ? httpsPort : httpPort;
+    const mediationLayer = wsConfig.mediationLayer;
+    const zoweInstance = process.env.ZOWE_INSTANCE || 'default';
+    const sessionCookieName = mediationLayer.enabled ? `connect.sid.${zoweInstance}` : `connect.sid.${cookiePort}`;
```
1000TurquoisePogs, line 194: I don't know if people currently customize ZOWE_INSTANCE, even though they should, so there are many that are ZOWE_INSTANCE=1. Is there a better way to share a cookie name for a cluster without it being the instance?

lib/index.js
```diff
@@ -330,6 +339,22 @@ Server.prototype = {
         eurekaOverrides: apimlConfig.eureka
       });
       yield this.apiml.registerMainServerInstance();
+      yield this.apiml.takeOutOfService();
+      const instanceId = this.apiml.getInstanceId();
+      let appServerClusterSize = +process.env.ZOWE_APP_SERVER_CLUSTER_SIZE;
+      if (!Number.isInteger(appServerClusterSize) || appServerClusterSize < 3) {
+        appServerClusterSize = 3;
+      }
```
1000TurquoisePogs, line 347: a code comment is needed in this section
1. why does it take out of service as soon as possible
2. why is the minimum cluster size 3?
I think the answer is in another file I haven't read yet, so a code comment will help people to understand.

lib/sync.ts
```diff
@@ -0,0 +1,62 @@
...
+  const data = { sid, session };
+  const sessionLogEntry: SessionSyncCommand = { type: 'session', payload: data };
+  syncLog.info(`updateSession log entry ${JSON.stringify(sessionLogEntry)}`);
+  syncEventEmitter.emit('session', sessionLogEntry);
+}
+
+export function setAllStorageForPlugin(pluginId: string, dict: KeyVal): void {
+  const action: StorageActionSetAll = { type: 'set-all', data: { pluginId, dict } };
+  const storageLogEntry: StorageSyncCommand = { type: 'storage', payload: action };
+  syncLog.info(`setAllStorageForPlugin log entry ${JSON.stringify(storageLogEntry)}`);
```
1000TurquoisePogs, line 29: when this gets merged, this should be debug or not existing, because there might be private information.
Also, I'm not certain but   
`syncLog.debug('setAllStorageForPlugin log entry',storageLogEntry);`
might perform better than
`syncLog.debug(`setAllStorageForPlugin log entry ${JSON.stringify(storageLogEntry)}`);`
If the object is not converted to a string until logging occurs, since you do not want to do processing on a log message that will not be printed.

lchudinov: `syncLog.debug('setAllStorageForPlugin log entry',storageLogEntry);` outputs `setAllStorageForPlugin log entry [object Object]`.

I would comment out or hide under `if (this.debug) { ... }`. The messages help only I need a deep debugging.

lib/index.js
lchudinov: All instances must be no-op until Leader is elected. 3 is a minimal cluster size. With 3 peers we can tolerate 1 failure.

lib/index.js
lchudinov: For Eureka statuses I introduced constants.

lchudinov: As for cookie name: maybe we can go with `connect.sid`. @1000TurquoisePogs  What do you think?

1000TurquoisePogs: This is ok for now.

lib/raft-rpc-ws.ts
```diff
@@ -0,0 +1,301 @@
...
+    this.pendingRequests.clear();
+  }
+
+  private onError(ws: WebSocket, err: Error): void {
+    raftLog.trace(`connection error ${JSON.stringify(err)}`);
+  }
+
+}
+
+export class RaftRPCWebSocketService {
```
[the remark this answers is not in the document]
cleonty: The Driver sends RPC requests. The Service receives RPC requests.  The Service is used here: https://github.com/zowe/zlux-server-framework/blob/6a2c346799d2caee4d388ce5775fa4d29ef9f320/lib/webapp.js#L39

lib/index.js
```diff
@@ -186,6 +184,9 @@ Server.prototype = {
     const httpPort = wsConfig.http ? wsConfig.http.port : undefined;
     const httpsPort = wsConfig.https ? wsConfig.https.port : undefined;
     const cookiePort = httpsPort ? httpsPort : httpPort;
+    const mediationLayer = wsConfig.mediationLayer;
+    const zoweInstance = process.env.ZOWE_INSTANCE || 'default';
+    const sessionCookieName = mediationLayer.enabled ? `connect.sid.${zoweInstance}` : `connect.sid.${cookiePort}`;
```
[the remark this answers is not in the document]
cleonty: Should be fixed now.

lib/sync-service.ts
```diff
@@ -0,0 +1,64 @@
...
+  constructor(
+    private raft: Raft,
+  ) {
+    this.init();
+  }
+
+  private init(): void {
+    const sessionChangeListener = this.onSessionChange.bind(this);
+    const storageChangeListener = this.onStorageChange.bind(this);
+    this.raft.stateEmitter.on('state', (state: State) => {
```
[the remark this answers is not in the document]
cleonty: During the election Raft ignores all changes, and it should be ok because API Gateway routes traffic only to a leader. I mean, none of instances receive traffic during Leader Election.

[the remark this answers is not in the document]
lchudinov: During the election Raft ignores all changes, and it should be ok because API Gateway routes traffic only to a leader. I mean, none of instances receive traffic during Leader Election.

lib/raft.ts
```diff
@@ -0,0 +1,1381 @@
...
+import * as fs from 'fs';
+import * as path from 'path';
+import { Request, Response } from "express";
+import { NextFunction } from "connect";
+import { StorageDict } from "./clusterManager";
+import { SyncService } from "./sync-service";
+const zluxUtil = require('./util');
+const raftLog = zluxUtil.loggers.raftLogger;
+
+// (global as any).COM_RS_COMMON_LOGGER.setLogLevelForComponentName("_zsf.raft", 4);
```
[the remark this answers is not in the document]
lchudinov: Fixed.

lib/index.js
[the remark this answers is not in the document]
lchudinov: Should be fixed now.

lib/raft-rpc-ws.ts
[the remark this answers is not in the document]
lchudinov: The Driver sends RPC requests. The Service receives RPC requests.  The Service is used here: https://github.com/zowe/zlux-server-framework/blob/6a2c346799d2caee4d388ce5775fa4d29ef9f320/lib/webapp.js#L39

lib/raft.ts
```diff
@@ -0,0 +1,1381 @@
...
+  constructor(
+    private stateFilename: string,
+    private snapshotFilename: string,
+  ) {
+    raftLog.debug(`raft state file: ${stateFilename}`);
+  }
+
+  saveState(state: string): void {
+    try {
+      fs.writeFileSync(this.stateFilename, state, 'utf-8');
```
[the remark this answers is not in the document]
lchudinov: There are two persisters: `FilePersister` and `DummyPersister`. We agreed that we should not use persistence at all so only DummyPersister is used. I agree that `FilePersister` should be fixed if we  decide to use it.

lib/raft.ts
```diff
@@ -0,0 +1,1381 @@
...
+    if (this.state !== 'Candidate') {
+      this.state = 'Candidate';
+      this.emitState();
+    }
+    this.currentTerm++;
+    this.votedFor = this.me;
+    let votes = 1;
+    let done = false;
+    const term = this.currentTerm;
+    const peerCount = this.peers.length;
```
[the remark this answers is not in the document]
lchudinov: Cluster membership change is not addressed yet. Number of peers in the cluster is fixed for now. Raft gets the number of participants in the cluster using environment variable.

lib/raft.ts
```diff
@@ -0,0 +1,1381 @@
...
+        }
+        votes++;
+        if (done) {
+          this.tracePrintf("got vote from peer %s but election already finished", peerAddress);
+          return;
+        } else if (this.state == 'Follower') {
+          this.tracePrintf("got heartbeat, stop election")
+          done = true;
+          return;
+        } else if (votes <= Math.floor(peerCount / 2)) {
```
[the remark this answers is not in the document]
lchudinov: Number of participants in the cluster is fixed for now.

lib/raft.ts
```diff
@@ -0,0 +1,1381 @@
...
+    }
+    this.heartbeatTimeoutId = setTimeout(() => this.sendHeartbeat(), this.heartbeatInterval)
+  }
+
+  private adjustNextIndexForServer(server: number, conflict: Conflict) {
+    if (conflict.conflictIndex === -1 && conflict.conflictTerm === -1) {
+      if (conflict.logLength === 0 && this.lastSnapshot) {
+        this.tracePrintf("follower's log is empty(have it re-started?) and there is a snapshot, send the snapshot to the follower");
+        setImmediate(() => this.installSnapshotForServer(server, this.currentTerm, this.lastSnapshot));
+      } else {
```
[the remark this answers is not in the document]
lchudinov: It means the follower has just restarted so it has empty log, and the leader doesn't have a snapshot yet because there are not enough entries in the raft log. It this cases the leader will send log entries to the follower starting from beginning.

lib/raft.ts
```diff
@@ -0,0 +1,1381 @@
...
+        this.nextIndex[server] = conflict.logLength;
+        this.tracePrintf("set nextIndex for server %d = %d because there are missing entries in follower's log", server, this.nextIndex[server]);
+      }
+    } else if (conflict.conflictIndex !== -1) {
+      this.nextIndex[server] = conflict.conflictIndex;
+      this.tracePrintf("set nextIndex for server %d = %d because conflictIndex given", server, this.nextIndex[server]);
+    } else {
```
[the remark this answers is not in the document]
lchudinov: `conflictTerm` is an extra information that can be helpful. For now, only `conflitTerm === -1` matters.

lib/raft.ts
```diff
@@ -0,0 +1,1381 @@
...
+    const entries: RaftLogEntry[] = [];
+    let last = this.len();
+    if (kind == "appendentries") {
+      last = this.commitIndex + 1;
+    } else {
+      last = last - 1;
+    }
+    let start = this.nextIndex[server];
+    if (start < this.startIndex) {
+      start = this.startIndex;
```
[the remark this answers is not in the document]
lchudinov: Just in case, `nextIndex` must not became wrong.

lib/raft.ts
```diff
@@ -0,0 +1,1381 @@
...
+      console.log(`state is ${JSON.stringify(buffer.toString())}`);
+      return buffer.toString();
+    } catch (e) {
+      raftLog.warn(`unable to read raft persistent state: ${e}`, JSON.stringify(e));
+    }
+  }
+
+  readSnapshot(): string | undefined {
+    try {
+      const buffer = fs.readFileSync(this.snapshotFilename);
```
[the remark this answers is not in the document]
lchudinov: It was used in my experiments with persistence. I agree that `FilePersister` should be fixed if we decide to use it.

lib/raft.ts
```diff
@@ -0,0 +1,1381 @@
...
+  private peers: RaftPeer[]; // RPC end points of all peers
+  private me: number;  // this peer's index into peers[]
+  private state: State = 'Follower';
+  private readonly electionTimeout = Math.floor(Math.random() * (maxElectionTimeout - minElectionTimeout) + minElectionTimeout);
+  private trace = true;
+  private started = false;
+
+  // persistent state
+  private currentTerm: number = 0;
+  private votedFor = -1
```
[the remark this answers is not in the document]
lchudinov: Leader stops being a leader only if it observes another Leader with a higher term number.

lib/raft.ts
```diff
@@ -0,0 +1,1381 @@
...
+
+  // volatile state on all servers
+  private commitIndex: number = -1;
+  private lastApplied: number = -1;
+
+  // volatile state on leaders(Reinitialized after election):
+  private nextIndex: number[] = [];  //  for each server, index of the next log entry to send to that server (initialized to leader last log index + 1)
+  private matchIndex: number[] = []; // for each server, index of highest log entry known to be replicated on server (initialized to 0, increases monotonically)
+  private electionTimeoutId: NodeJS.Timer;
+  private readonly heartbeatInterval: number = Math.round(minElectionTimeout * .75);
```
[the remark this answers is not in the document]
lchudinov: Yes, it must be configurable along with election timeouts.

lib/raft.ts
```diff
@@ -0,0 +1,1381 @@
...
+  private lastSnapshot: Snapshot;
+
+  private persister: Persister;
+  private maxLogSize: number = -1;
+  private syncService: SyncService;
+  private apiml: ApimlConnector;
+
+
+  constructor() {
+    const clusterEnabledEnvVar = process.env.ZOWE_APP_SERVER_CLUSTER_ENABLED;
```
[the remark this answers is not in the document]
lchudinov: I renamed the variable to `ZLUX_RAFT_CLUSTER_ENABLED`

lib/raft.ts
```diff
@@ -0,0 +1,1381 @@
...
+    this.isEnabled = clusterEnabledEnvVar === 'TRUE' || clusterEnabledEnvVar === 'YES';
+  }
+
+  async start(apiml: ApimlConnector): Promise<void> {
+    raftLog.info(`Starting peer electionTimeout ${this.electionTimeout} ms heartbeatInterval ${this.heartbeatInterval} ms`);
+    this.apiml = apiml;
+    this.persister = Raft.makePersister();
+    this.maxLogSize = Raft.getMaxLogSize();
+    const { peers, me } = await this.waitUntilZluxClusterIsReady();
```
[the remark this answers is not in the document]
lchudinov: I renamed it to `waitUntilRaftClusterIsReady()`

lib/raft.ts
```diff
@@ -0,0 +1,1381 @@
...
+    this.peers = peers;
+    this.me = me;
+    if (me === -1) {
+      raftLog.warn(`unable to find my instance among registered zlux instances`);
+      return;
```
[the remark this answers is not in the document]
lchudinov: It is just an extra check. It must never happen unless the is a bug in APIML. If the instance has registered in Eureka, Eureka must include the instance into the list list of registered instances, otherwise it is a bug.

lib/raft.ts
```diff
@@ -0,0 +1,1381 @@
...
+    const me = zluxInstances.findIndex(instance => instance.instanceId === instanceId);
+    const peers = zluxInstances.map(instance => RaftPeer.make(instance, this.apiml));
+    return { peers, me };
+  }
+
+  private static makePersister(): Persister {
+    let persister: Persister;
+    if (process.env.ZLUX_RAFT_PERSISTENCE_ENABLED === "TRUE") {
+      raftLog.debug("raft persistence enabled");
+      let logPath = process.env.ZLUX_LOG_PATH!;
```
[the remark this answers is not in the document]
lchudinov: `!` at the end means I expect that variable is not `undefined`. I changed the location to `$INSTANCE_DIR/workspace/app-server`

1000TurquoisePogs:
Various odd behavior seen:

1. The description is wrong. the variable for enabling is ZOWE_APP_SERVER_CLUSTER_ENABLED

2. When enabled, The first RAFT message I see is

2020-06-23 01:28:39.032 <ZWED:10522> sgrady INFO (_zsf.raft,raft.js:256) zlux app-server cluster is ready, size 3, instances [0] = 10.117.112.103:zlux:12360(me), [1] = 10.117.112.103:zlux:12361, [2] = 10.117.112.103:zlux:12362
2020-06-23 01:29:07.663 <ZWED:10522> sgrady INFO (_zsf.raft,raft.js:369) ZWED0294I - Got final vote from wss://10.117.112.103:12361/raft and became leader of term 15

Is that normal to start at term 15? That seems weird.

3. When I close the primary server and a secondary wins election, the existing webpage breaks for some time. This seems to be related to apiml understanding to take into & out of service. I have an old apiml. Is this improved on newer versions?


Generally this works, because the behavior without RAFT is if I use 3 servers, they dont communicate and so you get random 401 when APIML cycles between the servers. With 1 server, a crash kills everything. 3 with RAFT is best so far.

lchudinov:
@1000TurquoisePogs, my answers below:

> 1. The description is wrong. the variable for enabling is ZOWE_APP_SERVER_CLUSTER_ENABLED
The variable name is `ZLUX_RAFT_CLUSTER_ENABLED`. It looks like 1) you haven't got my latest changes or 2) haven't re-built. Please update your source code and re-build.

> 2. When enabled, The first RAFT message I see is
> 
> 2020-06-23 01:28:39.032 ZWED:10522 sgrady INFO (_zsf.raft,raft.js:256) zlux app-server cluster is ready, size 3, instances [0] = 10.117.112.103:zlux:12360(me), [1] = 10.117.112.103:zlux:12361, [2] = 10.117.112.103:zlux:12362
> 2020-06-23 01:29:07.663 ZWED:10522 sgrady INFO (_zsf.raft,raft.js:369) ZWED0294I - Got final vote from wss://10.117.112.103:12361/raft and became leader of term 15
> 
> Is that normal to start at term 15? That seems weird.

It means there were 14 unsuccessful leader elections during the time from 01:28:39.032 to 01:29:07.663. Why the peers didn't send votes during that period of time? It looks like they were registered in Eureka, but Raft wasn't enabled. They sent votes as soon as you restarted them with Raft enabled. 

> 1. When I close the primary server and a secondary wins election, the existing webpage breaks for some time. This seems to be related to apiml understanding to take into & out of service. I have an old apiml. Is this improved on newer versions?

What do you mean saying "the existing page breaks"? It takes some time for API Gateway while it switches over to a new leader.  During that period HTTP requests may take longer than usually, but you shouldn't get bad responses. New APIML version must work better. I'm testing it.

1000TurquoisePogs:
What I meant by the page breaks is that during the time that the gateway was switching, HTTP requests responded with HTTP 500, rather than waiting. I was using 1.10 code, so updating may help.

lchudinov:
> What I meant by the page breaks is that during the time that the gateway was switching, HTTP requests responded with HTTP 500, rather than waiting. I was using 1.10 code, so updating may help.
@1000TurquoisePogs, what was body of the HTTP 500 response? Was it "RequestContextNotPreparedException" or something else?

1000TurquoisePogs:
Because this is not going to be needed for the foreseeable future, I'm closing it to keep the list of PRs short.

PR #207, opened 2020-05-01
Added node server to retry if initial connection to the server fails

sakeerthy:

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -214,7 +214,7 @@ Server.prototype = {
         ((this.startUpConfig.proxiedHost !== undefined) || (this.startUpConfig.proxiedPort !== undefined))) {
       const host = this.startUpConfig.proxiedHost;
       const port = this.startUpConfig.proxiedPort;
-      yield checkProxiedHost(host, port);
+      yield checkProxiedHost(host, port, this.userConfig.agent.handshakeTimeout);
     }
     this.webApp = makeWebApp(webAppOptions);
     yield this.webServer.startListening(this.webApp);
--- a/lib/proxy.js
+++ b/lib/proxy.js
@@ -24,6 +24,8 @@ const WebSocket = require('ws');
 const net = require('net');
 
 const proxyLog = util.loggers.proxyLogger;
+const RECONNECT_DELAY = 5000;
+const DEFAULT_HANDSHAKE_TIMEOUT = 30000;
 
 function convertOptions(request, realHost, realPort, urlPrefix) {
   var options = {};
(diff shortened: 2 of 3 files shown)


lib/index.js
```diff
@@ -214,7 +214,7 @@ Server.prototype = {
         ((this.startUpConfig.proxiedHost !== undefined) || (this.startUpConfig.proxiedPort !== undefined))) {
       const host = this.startUpConfig.proxiedHost;
       const port = this.startUpConfig.proxiedPort;
-      yield checkProxiedHost(host, port);
+      yield checkProxiedHost(host, port, this.userConfig);
```
1000TurquoisePogs, line 217: instead of passing in the whole config object, just pass in the parameter. because then we chance its location over time without effecting the function itself

lib/proxy.js
```diff
@@ -301,24 +301,40 @@ function makeWsProxy(host, port, urlPrefix, isHttps) {
   };
 };
 
-function checkProxiedHost(host, port) {
+function checkProxiedHost(host, port, userConfig) {
   const client = new net.Socket();
+  const reconnectDelay = 5000;
```
1000TurquoisePogs, line 306: Should this be a const inside or outside the function? Usually, values that we want to be able to change in the future would be defined as a const at the top of the file. I guess it depends

lib/proxy.js
```diff
@@ -301,24 +301,40 @@ function makeWsProxy(host, port, urlPrefix, isHttps) {
...
+        if (host === '127.0.0.1') {
+          proxyLog.warn("ZWED0046W"); //proxyLog.warn("The auth services host system was not " +
           //"specified at startup, and defaulted to 127.0.0.1.\n" +
           //"Verify that the auth services server is running, " +
           //"or specify at startup the remote host and port to connect to. " +
           //"See documentation for details.");
+        }
+        reject(`Communication with ${host}:${port} failed: ${e.toString()}`);
+      }
+      if (e.code == 'ECONNREFUSED') {
```
1000TurquoisePogs, line 331: This needs to be else-if

lib/proxy.js
```diff
@@ -301,24 +301,40 @@ function makeWsProxy(host, port, urlPrefix, isHttps) {
   };
 };
 
-function checkProxiedHost(host, port) {
+function checkProxiedHost(host, port, userConfig) {
   const client = new net.Socket();
+  const reconnectDelay = 5000;
+  let timeLeft = userConfig.agent.handshakeTimeout;
```
1000TurquoisePogs, line 307: const DEFAULT_HANDSHAKE_TIMEOUT=30000
userConfig.agent.handshakeTimeout ? userConfig.agent.handshakeTimeout : DEFAULT_HANDSHAKE_TIMEOUT

PR #215, opened 2020-05-18
Speedup for connecting to ssh with node12+

1000TurquoisePogs:
dh creation is much slower on node 12 than prior. Use sw library until a better solution found. It fulfills the same api, but is faster than what node12 is doing.


1000TurquoisePogs:
Actually no need to merge to master, just going to merge to staging.

PR #226, opened 2020-06-15
Node support warn

DivergentEuropeans:
Created node support warn message + timetable for ez plug & play of version number

### DIFF
--- a/lib/webapp.js
+++ b/lib/webapp.js
@@ -55,6 +55,9 @@ const DEFAULT_RELOAD_TIMEOUT = 2000; //2 seconds
 const DEFAULT_HSTS_TIME_SECONDS = 604800; //1 week
 const nodeVer = process.version.substring(1, process.version.length);
 let nodeMajorVer = nodeVer.split('.')[0];
+const nodeSupportTimeTable = { // Signifies end of support
+  6: new Date('2020-09-1'), // Order by highest vers first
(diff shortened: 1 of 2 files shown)


lib/webapp.js
```diff
@@ -187,6 +192,32 @@ function createDataserviceStorage(pluginId) {
...
+  for (var majorVer in nodeSupportTimeTable) {
+    var months = nodeSupportTimeTable[majorVer].getMonth() - currentDate.getMonth() + 
+      (12 * (nodeSupportTimeTable[majorVer].getFullYear() - currentDate.getFullYear()));
+
+    if (parseInt(nodeMajorVer) <= majorVer) {
+      if (currentDate > nodeSupportTimeTable[majorVer]) // We use break so if version is multiple iterations old, we don't get multiple messages
+      {
+        logger.warn("YOUR NODE VERSION IS OUT OF DATE. ZOWE IS NO LONGER SUPPORTING NODE VERSION " + majorVer + ".");
+        break;
+      } else if (months >= 6) {
```
timgerstel, line 207: This seems like unnecessarily verbose logging that may need to be cleaned up later when its irrelevant (after node version 6 is unsupported)

DivergentEuropeans:
Realized after the fact this should be done via node checking script in install-packaging

PR #228, opened 2020-06-15
Add TypeScript Conversion Plugin Loader

ParthS007:


lib/plugin-loader.ts
```diff
@@ -0,0 +1,853 @@
...
+    //      bootstrapLogger.warn("Authentication plugin was found which was not requested in "
+    //          + "the server configuration file's dataserviceAuthentication object. "
+    //          + "Skipping load of this plugin");
+    //      return false;
+    //    }
+    return true;
+  }
+}
+
+const plugInConstructorsByType = {
```
1000TurquoisePogs, line 500: This is very similar in concept to an enum, so maybe it should be made into one.

lib/plugin-loader.ts
```diff
@@ -0,0 +1,853 @@
...
+  "library": LibraryPlugIn,
+  "application": ApplicationPlugIn,
+  "windowManager": WindowManagerPlugIn,
+  "bootstrap": BootstrapPlugIn,
+  "desktop": DesktopPlugIn,
+  "nodeAuthentication": NodeAuthenticationPlugIn,
+  "proxyConnector": ProxyConnectorPlugIn
+};
+
+export class makePlugin {
```
1000TurquoisePogs, line 510: Not all functions are should be classes. This one appears to have been a factory function, which should have remained a function.

lib/plugin-loader.ts
```diff
@@ -0,0 +1,853 @@
...
+              error: new Error(`ZWED0018E - No plugin type found, skipping`)};
+    }
+    bootstrapLogger.info(`ZWED0214I`, pluginBasePath, pluginDef.identifier, pluginDef.pluginType); //bootstrapLogger.info(`Read ${pluginBasePath}: found plugin id = ${pluginDef.identifier}, `
+        //+ `type = ${pluginDef.pluginType}`);
+    pluginDef.location = pluginBasePath;
+    return pluginDef;
+  }
+
+  private _readPluginDefAsync(pluginDescriptorFilename: any) {
+    return new Promise((resolve, reject)=> {
```
1000TurquoisePogs, line 600: It's especially important to show the return type when the return type is a promise.
`private _readPluginDefAsync(pluginDescriptorFilename: any): Promise<any>` or `private _readPluginDefAsync(pluginDescriptorFilename: any): Promise<object>`  seems correct here.

lib/plugin-loader.ts
```diff
@@ -0,0 +1,853 @@
...
+    zluxUtil.deepFreeze(plugin);
+    this.plugins.push(plugin);
+    this.pluginMap[plugin.identifier] = plugin;
+    this.emit('pluginAdded', {
+      data: plugin
+    });
+  }
+};
+
+module.exports = PluginLoader;
```
1000TurquoisePogs, line 841: Believe it or not, you do not need to use `module.exports` in typescript. The reason why is because the `import` and `export` system compiles into `module.exports` in the case that the compile type is nodejs. `export` would do something different if compiled for a browser. there are a few choices.

lib/plugin-loader.ts
```diff
@@ -0,0 +1,853 @@
...
+      var osDelim = operatingSystem === "win32" ? ";" : ":"; 
+      bootstrapLogger.info("ZWED0116I", serverModuleLocation, clientModuleLocation); //bootstrapLogger.log(bootstrapLogger.INFO,
+        //`The LOCATIONS are ${serverModuleLocation} and ${clientModuleLocation}`);
+      nodePathAdd = `${process.env.NODE_PATH}${serverModuleLocation}${osDelim}${clientModuleLocation}${osDelim}.`
+      process.env.NODE_PATH = nodePathAdd;
+      require("module").Module._initPaths();    
+      bootstrapLogger.info("ZWED0117I", fileLocation); //bootstrapLogger.log(bootstrapLogger.INFO, `The fileLocation is ${fileLocation}`);
+      bootstrapLogger.info("ZWED0118I", process.env.NODE_PATH); //bootstrapLogger.log(bootstrapLogger.INFO,
+        //`The NODE_PATH is ${process.env.NODE_PATH}`);
+      const nodeModule = require(fileLocation);
```
1000TurquoisePogs, line 131: require is often similar to typescript's `import`, except in this case. most require occur at the top of the file: a static require/import. I think import only works statically, while you can require dynamically by having require inside a function like we do here. But, typescript might give a compile warning, so you might want to do:
`const nodeModule = (require as any)(fileLocation);`
typecasting to `as any` will make annoying typescript messages disappear sometimes.

1000TurquoisePogs (changes requested):
hope that this exercise helped you to understand typescript more. the PR contains bugs, so please look at my comments to learn more

lib/plugin-loader.ts
```diff
@@ -0,0 +1,853 @@
...
+
+export class makeDataService extends Service {
+  private context: any;
+
+  constructor(configuration: any, plugin: any, def: any) {
+    super(configuration, plugin, def)
+    configuration = configService.getServiceConfiguration(this.plugin.identifier, this.plugin.location,
+      this.def.name, this.context.config, this.context.productCode);
+    let dataservice: any;
+    if (this.def.type == "external") {
```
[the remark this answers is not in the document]
ParthS007: I have added enum of possible types. can you please check now if this fine.

lib/plugin-loader.ts
ParthS007: Can you please check now, there were some errors in below `makePlugin` after adding enum.

PR #229, opened 2020-06-24
Bugfix for plugin default config and terminal handler location

1000TurquoisePogs:
Fixes to reduce need to copy files during installation, and to bugfix an edge case where there was no instance_dir equivalent to a root_dir setting file

Previously, terminal handlers could only exist within ROOT_DIR. This was just overlooked during CUPIDS work, now it can live in the workspace dir of the plugin within the app-server.

Previously, serverside plugin config had to exist within the instance directory, meaning plugins had to perform a copy operation during installation. Now, this should be able to exist within the plugins own folder, like other configuration files.

### DIFF
--- a/plugins/terminal-proxy/lib/terminalProxy.js
+++ b/plugins/terminal-proxy/lib/terminalProxy.js
@@ -16,6 +16,7 @@ var net = require('net');
 const tls = require('tls');
 const crypto = require('crypto');
 const fs = require('fs');
+const pathModule = require('path');
 const ssh = require('./ssh');
 const SSH_MESSAGE = ssh.MESSAGE;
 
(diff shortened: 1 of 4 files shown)


plugins/terminal-proxy/lib/terminalProxy.js
```diff
@@ -701,22 +702,49 @@ var handlerModules = null;
...
   if (handlerModules == null) {
     handlerModules = [];
-    let filenames = fs.readdirSync(__dirname);
-    let len = filenames.length;
+    let handlers = {}; //key: filename, value: path
+
+    let filenames;
+    let len;
+    try {
+      let handlerDir = pathModule.join(process.env.INSTANCE_DIR,'workspace','app-server','org.zowe.terminal.proxy','handlers');
```
DivergentEuropeans, line 710: Do you need to check if process.env.INSTANCE_DIR exists?

PR #231, opened 2020-06-30
Bugfix for zss refresh not being issued

1000TurquoisePogs:
Fix for regression where session would expire prematurely because app server would not request a refresh from ZSS

The problem: if apiml was being used, plugin would report refresh isnt supported, even though it's partially supported. This meant that the session would expire when the shorted session timer (zss, 1 hour) expired, rather than allowing it to renew until hitting the non-expiring, longer apiml one.

This issue did not occur with sso, but did occur when using non-sso apiml+zss.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 All notable changes to the Zlux Server Framework package will be documented in this file.
 This repo is part of the app-server Zowe Component, and the change logs here may appear on Zowe.org in that section.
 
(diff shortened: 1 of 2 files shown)


NakulManchanda:
[zowe build](https://zowe.jfrog.io/zowe/webapp/#/builds/zowe-install-packaging%20::%20serve-compressed/7/1593618527497) with pr 231 zlux-server-framework and pr 247 zlux-app-manager

PR #235, opened 2020-07-10
Load recognizers & actions from plugins & merge with config location

DivergentEuropeans:
Load recognizers & actions from plugins & merge with config location


utils/install-app.js
```diff
@@ -139,6 +141,88 @@ function addToServer(appDir, installDir) {
...
+  try {
+    recognizers = JSON.parse(fs.readFileSync(path.join(appDir, "config/recognizers", appId))).recognizers;
+    const keys = Object.keys(recognizers)
+    for (const key of keys) {
+      recognizers[key].pluginVersion = appVers;
+      recognizers[key].pluginIdentifier = appId;
+    }
+    logger.debug('ZWED0294I', appId); //logger.debug("Found recognizers for '" + appId + "'");
+  } catch (e) {
+    logger.debug('ZWED0295I', (path.join(appDir, "config/recognizers", appId)), appId); //logger.debug("Could not find recognizers in '" + (path.join(appDir, "config/recognizers", appId)) + "'");
```
DivergentEuropeans, line 158: Here we get config & recognizers for the ones existing in a plugin (/appDir/config/xxx)

utils/install-app.js
```diff
@@ -139,6 +141,78 @@ function addToServer(appDir, installDir) {
...
+  }
+
+  try { // Get pre-existing recognizers in config, if any
+    configLocation = path.join(process.env.ZLUX_CONFIG_FILE.substring(1), "../../ZLUX/pluginStorage/org.zowe.zlux.ng2desktop/");
```
1000TurquoisePogs, line 162: I'm pretty sure this env var is NOT guaranteed to exist.
when the server runs, INSTANCE_DIR is guaranteed.
When run from install-app.sh, it should be also, but it look like it isnt... easy fix? It seems like a minor refactor is needed because when INSTANCE_DIR doesnt exist, the user at least has provided the server.json location: 
https://github.com/zowe/zlux-app-server/blob/master/bin/install-app.sh#L74
... it ends up as "zluxConfig" in the top of this file.

utils/install-app.js
```diff
@@ -139,6 +141,78 @@ function addToServer(appDir, installDir) {
...
 }
 
+function loadRecognizers(appDir, appId, appVers) {
+  let recognizers;
+  let configRecognizers;
+  let configLocation;
+
+  try { // Get recognizers in a plugin's appDir/config/xxx location
+    recognizers = JSON.parse(fs.readFileSync(path.join(appDir, "config/recognizers", appId))).recognizers;
+    const keys = Object.keys(recognizers)
```
1000TurquoisePogs, line 151: In the example on your other PR, recognizers is an array, not an object.
{
  "recognizers":[
    {
      "id":"com.rs.iuw.mq.openapp.forward",

utils/install-app.js
```diff
@@ -139,6 +141,78 @@ function addToServer(appDir, installDir) {
...
+    }
+    logger.debug('ZWED0294I', appId); //logger.debug("Found recognizers for '" + appId + "'");
+  } catch (e) {
+    logger.debug('ZWED0295I', (path.join(appDir, "config/recognizers", appId)), appId); //logger.debug("Could not find recognizers in '" + (path.join(appDir, "config/recognizers", appId)) + "'");
+  }
+
+  try { // Get pre-existing recognizers in config, if any
+    configLocation = path.join(process.env.ZLUX_CONFIG_FILE.substring(1), "../../ZLUX/pluginStorage/org.zowe.zlux.ng2desktop/");
+    configRecognizers = JSON.parse(fs.readFileSync(path.join(configLocation, "recognizers", appId))).recognizers;
+    recognizers = Object.assign(configRecognizers, recognizers); // // If found, combine the ones found in config with ones found in plugin
```
1000TurquoisePogs, line 164: If your key for the object (looked like an array, is this correct???) is not meaningfully unique, then object.assign could easily end up with duplicates.

If it's an array, then you need to iterate through and make sure there is no copy based on pluginversion, pluginidentifier, and id.

If it's not an array, then the key could just be pluginid:pluginversion:id

utils/install-app.js
DivergentEuropeans: Yes, it is. It's an array of objects so this is a good way to traverse it. I could also use .map if I wanted to

The rest of searches assume that the elements are 0 through n which I didn't want to do, because it's an unnecessary assumption imo

DivergentEuropeans: I didn't use Array.keys because it's not supported in IE

utils/install-app.js
DivergentEuropeans: I agree, merging algorithm needs to be better

utils/install-app.js
```diff
@@ -141,26 +141,41 @@ function addToServer(appDir, installDir) {
...
     }
     logger.debug('ZWED0294I', appId); //logger.debug("Found recognizers for '" + appId + "'");
   } catch (e) {
     logger.debug('ZWED0295I', (path.join(appDir, "config/recognizers", appId)), appId); //logger.debug("Could not find recognizers in '" + (path.join(appDir, "config/recognizers", appId)) + "'");
   }
 
+  if (process.env.INSTANCE_DIR) {
+    configLocation = path.join(process.env.INSTANCE_DIR, "workspace/app-server/ZLUX/pluginStorage/org.zowe.zlux.ng2desktop/");
+  } else {
+    configLocation = path.join(userInput.zluxConfig, "../../ZLUX/pluginStorage/org.zowe.zlux.ng2desktop/");
```
1000TurquoisePogs, line 166: zluxConfig should be read to find the value of the instance directory, rather than assuming it.
It would be `instanceDir`, meaning path.join(instanceDir, 'ZLUX/pluginStorage/org.zowe.zlux.ng2desktop/`);
see an example here https://github.com/zowe/zlux-app-server/blob/master/defaults/serverConfig/server.json#L43

utils/install-app.js
```diff
@@ -139,6 +141,107 @@ function addToServer(appDir, installDir) {
...
 }
 
+function copyRecognizers(appDir, appId, appVers) {
+  let recognizers;
+  let recognizersKeys;
+  let configRecognizers;
+  let configLocation;
+
+  try { // Get recognizers in a plugin's appDir/config/xxx location
+    recognizers = JSON.parse(fs.readFileSync(path.join(appDir, "config/recognizers", appId))).recognizers;
```
1000TurquoisePogs, line 151: you need to read all files in that folder, not just the one with the same name as your appId
App A can include recognizer files for both A and B.
Put this entire thing in a loop, and read each file you find.

utils/install-app.js
```diff
@@ -139,6 +141,107 @@ function addToServer(appDir, installDir) {
...
+}
+
+function copyActions(appDir, appId, appVers) {
+  let actions;
+  let actionsKeys;
+  let configActions;
+  let configLocation;
+
+  try { // Get actions in a plugin's appDir/config/xxx location
+    actions = JSON.parse(fs.readFileSync(path.join(appDir, "config/actions", appId))).actions;
```
1000TurquoisePogs, line 202: on the other hand this is perfectly fine. the app should only contain actions for itself, I think.

DivergentEuropeans:
I changed it to replacement for actions

PR #237, opened 2020-07-31
Present component and port data to web for dependency checks

1000TurquoisePogs:
Allow a subset of env info out to the browser for dependency checks and improved server-to-server communication
To get all environment details, rbac must be on and the person must effectively be an admin, but this change is for much less info for non-admins.

### DIFF
--- a/lib/webapp.js
+++ b/lib/webapp.js
@@ -245,25 +245,52 @@ const release = os.release();
 const cpus = os.cpus();
 const hostname = os.hostname();
 
-function getUserEnv(){
+function getUserEnv(rbac){
   var date = new Date();
   return new Promise(function(resolve, reject){
-    resolve({
-      "timestamp": date.toUTCString(),
-      "args": process.argv,
-      "nodeArgs": process.execArgv,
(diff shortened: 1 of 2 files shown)


lib/webapp.js
```diff
@@ -657,7 +691,7 @@ const staticHandlers = {
         return res.status(405).json({error: 'ZWED0129E - Only POST method supported'});
       });
       router.get('/environment', function(req, res){
-        getUserEnv().then(result => {
+        getUserEnv(rbac).then(result => {
```
DivergentEuropeans, line 694: I think you're effectively checking for rbac twice and duplicating code, and will have to do it this way for each subsequent endpoint that works w/ rbac and partially works w/o rbac.

What you could do instead is, use `router.get('/environment' [...]` **once** outside the main if/else, then leave the actual conditional endpoints to the conditional check

DivergentEuropeans (changes requested):
This works for me functionally, but it can probably be written better

DivergentEuropeans (approved):

PR #243, opened 2020-08-26
Plugin JSON attributes & requirements checking

DivergentEuropeans:
How to use? 

Change plugin definition to conform to new proposal in: https://github.com/zowe/zlux/wiki/ZLUX-Plugin-Definition-&-Structure#general-attributes

For example (in the Editor)

```
{
  "identifier": "org.zowe.editor",
  "apiVersion": "1.0.0",
  "pluginVersion": "2.2.1",
  "license": "EPL-2.0",
  "author": "Zowe",
  "homepage": "https://github.com/zowe/zlux-editor",
  "pluginType": "application",
  "webContent": {
    "framework": "angular",
    "launchDefinition": {
      "pluginShortNameKey": "editorTitle",
      "pluginShortNameDefault": "Editor",
      "imageSrc": "assets/icon.png"
    },
    "descriptionKey": "MyPluginDescription",
    "descriptionDefault": "Zowe Editor",
    "defaultWindowStyle": {
      "width": 1200,
      "height": 600
    }
  },
  "configurationData": {
    "resources": {
      "ui": {
        "aggregationPolicy": "override",
        "subResources": {
          "history": {
            "aggregationPolicy": "override"
          },
          "openTabs": {
            "aggregationPolicy": "override"
          }
        }
      }
    }
  },
  "requirements" : {
    "components": {
      "app-server": {
        "os": [
          "linux",
          "macos",
          "win32",
          "zos",
          "android"
        ],
        "cpu": [
  
        ]
      },
      "zss": {
        "version": [ ">=1.14.0" ],
        "os": [
          "zos"
        ],
        "cpu": [
          "OS/390"
        ],
        "endpoints": [
          "/datasetMetadata/**",
          "/datasetContents/**"
        ]
      }
    }
  }
}

```

### DIFF
--- a/lib/assets/i18n/log/messages_en.json
+++ b/lib/assets/i18n/log/messages_en.json
@@ -180,7 +180,10 @@
   "ZWED0293I":"Handling scan plugin request from worker=%d",
   "ZWED0294I":"Successfully loaded %d recognizers for '%s' into config",
   "ZWED0295I":"Successfully loaded %d actions for '%s' into config",
-  
+  "ZWED0296I": "(%s) is a supported platform for %s to install (%s).",
+  "ZWED0297I": "(%s) is a supported architecture for %s to install (%s).",
+  "ZWED0298I": "(%s) is a supported endpoint for %s to install (%s).",
+
   "ZWED0003W":"%s: Session security call %s failed for auth handler %s. Plugin response: %s",
   "ZWED0004W":"Tomcat for ID=%s not starting, no services succeeded loading",
   "ZWED0005W":"",
@@ -292,6 +295,8 @@
   "ZWED0170W":"Plugin (%s) loading failed. Message: \"%s\"",
   "ZWED0171W":"Rejected undefined referrer for url=%s, ip=%s",
   "ZWED0172W":"Rejected bad referrer=%s for url=%s, ip=%s",
+  "ZWED0173W":"%s could not verify (%s) as a supported platform to install (%s). Proceeding anyway...",
+  "ZWED0174W":"%s could not verify (%s) as a supported architecture to install (%s). Proceeding anyway...",
   
   "ZWED0001E":"RESERVED: Error: %s",
   "ZWED0002E":"Could not stop language manager for types=%s",
--- a/lib/assets/i18n/log/messages_en.json
+++ b/lib/assets/i18n/log/messages_en.json
@@ -295,13 +295,10 @@
   "ZWED0170W":"Plugin (%s) loading failed. Message: \"%s\"",
   "ZWED0171W":"Rejected undefined referrer for url=%s, ip=%s",
   "ZWED0172W":"Rejected bad referrer=%s for url=%s, ip=%s",
-<<<<<<< HEAD
-  "ZWED0173W":"%s could not verify (%s) as a supported platform to install (%s). Proceeding anyway...",
-  "ZWED0174W":"%s could not verify (%s) as a supported architecture to install (%s). Proceeding anyway...",
-=======
   "ZWED0173W":"Unable to decode P12 certificate (different password than keystore?). Attempting to use empty string as password. Decode error: %s.",
->>>>>>> 8c562d1b43aeea146672704db1c1062f77ecd656
-  
+  "ZWED0174W":"%s could not verify (%s) as a supported platform to install (%s). Proceeding anyway...",
+  "ZWED0175W":"%s could not verify (%s) as a supported architecture to install (%s). Proceeding anyway...",
+
   "ZWED0001E":"RESERVED: Error: %s",
   "ZWED0002E":"Could not stop language manager for types=%s",
   "ZWED0003E":"Loopback configuration not valid,%s\nLoopback calls will fail!",
--- a/lib/assets/i18n/log/messages_en.json
+++ b/lib/assets/i18n/log/messages_en.json
@@ -364,5 +364,8 @@
   "ZWED0149E":"SAF keyring reference missing userId \"%s\", keyringName \"%s\", or label \"%s\"",
   "ZWED0150E":"Cannot load SAF keyring due to missing keyring_js library",
   "ZWED0151E":"RESERVED: Env var %s not found",
-  "ZWED0152E":"Unable to locate server config instance location and INSTANCE_DIR environment variable does not exist."
+  "ZWED0152E":"Unable to locate server config instance location and INSTANCE_DIR environment variable does not exist.",
+  "ZWED0153E":"RESERVED: (%s) is not a supported platform for %s. Skipping (%s)... Supported: %s",
+  "ZWED0154E":"RESERVED: (%s) is not a supported architecture for %s. Skipping (%s)... Supported: %s",
+  "ZWED0155E":"RESERVED: (%s) is not a supported endpoint for %s. Skipping (%s)... Supported: %s"
 }
--- a/lib/index.js
+++ b/lib/index.js
@@ -299,7 +299,7 @@ Server.prototype = {
         });
       }
     }, installLogger));
-    this.pluginLoader.loadPlugins();
+    yield this.pluginLoader.loadPlugins();
     yield this.authManager.loadAuthenticators(this.userConfig, Object.assign({},this.webServer.getTlsOptions()));
     this.authManager.validateAuthPluginList();
     this.processManager.addCleanupFunction(function() {
--- a/lib/plugin-loader.js
+++ b/lib/plugin-loader.js
@@ -980,7 +980,7 @@ PluginLoader.prototype = {
                 if (requiredOS.includes(environmentComponent.os)) {
                   bootstrapLogger.debug("ZWED0296I", environmentComponent.os, name, plugin.identifier); //(`(${environmentComponent.os}) is a supported platform for ${name} to install (${plugin.identifier}).`);
                 } 
-                else if (acceptedOS.includes("!"+environmentComponent.os)) {
+                else if (requiredOS.includes("!"+environmentComponent.os)) {
                   plugin.error = new Error(`(${environmentComponent.os}) is not a supported platform for ${name}. Skipping (${plugin.identifier})... Supported: ${acceptedOS}`);
                 }
                 else {
--- a/lib/plugin-loader.js
+++ b/lib/plugin-loader.js
@@ -782,8 +782,9 @@ PluginLoader.prototype = {
     let newPlugins = [];
     let successCount = 0;
     const depgraph = new DependencyGraph(this.plugins);
-    envComps = await this.getComponentCapabilities(pluginContext.config); // Gets the environment and server information
-
+    if (Object.keys(envComps).length === 0) {
+      await this.getComponentCapabilities(pluginContext.config); // Gets the environment and server information
+    }
       for (const pluginDef of pluginDefs) {
         if (pluginDef.error) {
           //downstream will pick up error
@@ -951,10 +952,9 @@ PluginLoader.prototype = {
         /* TODO: before checking if dependencies are met, we must learn about the components that exist. doing this is not formalized
         currently, so we currently have a block per component to learn about their capabilities, version, environment, etc. perhaps in the 
         future zowe components could have metadata files and/or expected URLs for querying.*/
-        let obj = {};
-        obj[AGENT_COMP_ID] = agentComp;
-        obj[APP_SERVER_COMP_ID] = appServerComp;
-        return obj;
+        envComps[AGENT_COMP_ID] = agentComp;
+        envComps[APP_SERVER_COMP_ID] = appServerComp;
+        return;
       });
     });
   }
--- a/lib/plugin-loader.js
+++ b/lib/plugin-loader.js
@@ -981,7 +981,7 @@ PluginLoader.prototype = {
                   bootstrapLogger.debug("ZWED0296I", environmentComponent.os, name, plugin.identifier); //(`(${environmentComponent.os}) is a supported platform for ${name} to install (${plugin.identifier}).`);
                 } 
                 else if (requiredOS.includes("!"+environmentComponent.os)) {
-                  plugin.error = new Error(`(${environmentComponent.os}) is not a supported platform for ${name}. Skipping (${plugin.identifier})... Supported: ${acceptedOS}`);
+                  plugin.error = new Error(`(${environmentComponent.os}) is not a supported platform for ${name}. Skipping (${plugin.identifier})... Supported: ${requiredOS}`);
                 }
                 else {
                   bootstrapLogger.warn("ZWED0174W", name, environmentComponent.os, plugin.identifier); //`(${name} could not verify (${environmentComponent.os}) as a supported platform to install (${plugin.identifier}). Proceeding anyway...`);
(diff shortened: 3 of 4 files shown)


lib/plugin-loader.js
```diff
@@ -752,71 +756,275 @@ PluginLoader.prototype = {
...
+            } else if (acceptedCPU.includes("!"+components[AGENT_COMP_ID].cpu)) {
+              plugin.error = new Error(`(${components[AGENT_COMP_ID].cpu}) is not a supported architecture for the agent. Skipping (${plugin.identifier})... Supported: ${acceptedCPU}`);
+            }
+            else {
+              console.log(`Agent could not verify (${components[AGENT_COMP_ID].cpu}) as a supported architecture to install (${plugin.identifier}). Proceeding anyway...`);
+            }
+          }
+          if (neededEndpoints && neededEndpoints.length > 0 && components[AGENT_COMP_ID].endpoints) {
+            for (let i = 0; i < neededEndpoints.length; i++) {
+              if (!components[AGENT_COMP_ID].endpoints.includes(neededEndpoints[i])) {
```
lchudinov, line 1018: Actually, Agent's enpoints are urlMasks(e.g. `/datasetMetadata/**`) so you should check matches against URL masks.

lib/plugin-loader.js
```diff
@@ -752,71 +756,275 @@ PluginLoader.prototype = {
...
       }
     }
-    const sortedAndRejectedPlugins = depgraph.processImports();
-    sortedAndRejectedPlugins.plugins = sortedAndRejectedPlugins.plugins.filter((plugin)=> {
-      return !this.pluginMap[plugin.identifier];
+    if (host && port) {
+      let url = 'http://' + host + ':' + port + '/server/agent/environment';
+    
+      return new Promise((resolve, reject) => { /* Obtains and stores environment information from agent */
+        http.get(url, (res) => {
```
lchudinov, line 855: Please issue only a single HTTP request to obtain environment info, cache, and re-use the info when needed.

lib/plugin-loader.js
```diff
@@ -752,71 +756,275 @@ PluginLoader.prototype = {
...
+      }).on('error', (e) => {
+        bootstrapLogger.severe(e.message);
+        resolve();
+      });
+    }).then(() => { /* Obtains and stores the endpoints exposed by the agent */
+
+      'http://' + host + ':' + port + '/server/agent/services';
+
+      return new Promise((resolve, reject) => { 
+        http.get(url, (res) => {
```
lchudinov, line 893: Please issue only a single HTTP request to obtain services, cache, and re-use the info when needed.

lib/webapp.js
```diff
@@ -412,7 +412,9 @@ const staticHandlers = {
               loader.once('refreshFinish', (event) => {
                 respondToGetPlugins(req, res, webapp, type, webapp.plugins);
               });
-              loader.installPlugins(defs);
+              loader.installPlugins(defs).then( () => {
```
cleonty, line 415: Catch an error here.

lchudinov, line 415: Catch an error here.

lchudinov:
It is good for the beginning. I checked that it works. However, there is still a lot to do.

lib/plugin-loader.js
DivergentEuropeans: This request fires once and then the info gets checked each time per plugin

lib/plugin-loader.js
```diff
@@ -752,71 +756,275 @@ PluginLoader.prototype = {
...
-    for (const pluginDef of pluginDefs) {
-      if (pluginDef.error) {
-        //downstream will pick up error
-        newPlugins.push(pluginDef);
-      } else {
-        depgraph.addPlugin(pluginDef);
+    const componentCapabilities = await this.getComponentCapabilities(pluginContext.config); // Gets the environment and server information
+
+      for (const pluginDef of pluginDefs) {
+        if (pluginDef.error) {
```
timgerstel, line 762: This logic (762-772) could be refactored i believe

for ( pluginDef of pluginDefs) {
 this.checkPluginRequirements(pluginDef, componentCapabilities);
 if(pluginDef.error) {
  newPLugins....
 } else {
  deGraph...
 }
}

lib/plugin-loader.js
```diff
@@ -752,71 +756,275 @@ PluginLoader.prototype = {
...
+  async getComponentCapabilities(config) {
+    let appServerComp = {};
+    let agentComp = {};
+    let host;
+    let port;
+
+    appServerComp.os = process.platform; // Operating system
+    appServerComp.cpu = process.arch; // CPU architecture
+
+    if (config && config.agent.host) {
```
timgerstel, line 843: I think 843-848 can be refactored as well. If the app server started, you can assume it received a valid host and port for ZSS. The only instance where one wouldnt be set is if someone was using trivial auth, in which case this endpoint/functionality would be useless.

This just looks messy more so than anything, the above criticism could be a bit overstated

DivergentEuropeans:
Sean: I'm not a fan of "pluginReqs" because youre already in the plugin so its just "reqs"... we already made this mistake a bit with "pluginVersion" versus "apiVersion", where it could have just been "version" and "apiVersion"
1:29
reqs vs dependencies, dont really care, but prefixing with plugin is just not pretty

lib/plugin-loader.js
```diff
@@ -752,71 +756,275 @@ PluginLoader.prototype = {
...
-          this.plugins.push(frozen);
-          newPlugins.push(frozen);
-          this.pluginMap[plugin.identifier] = plugin;
-          successCount++;
-        } else {
-          bootstrapLogger.info("ZWED0125I", pluginDef.identifier); //bootstrapLogger.log(bootstrapLogger.INFO,
-            //`Plugin ${pluginDef.identifier} not loaded`);
-          this.pluginMap[pluginDef.identifier] = {}; //mark it as having failed so it will not be retried
+
+        if (acceptedComponents[AGENT_COMP_ID]) { /* --- Agent component --- */
```
1000TurquoisePogs, line 964: not seeing why you need a block per each component here. the claims are the same:
do os match
do cpu match
do version match
do endpoints match
and maybe more.
you could run through a loop rather than pasting a block per component to do these checks.

lib/plugin-loader.js
```diff
@@ -752,71 +756,275 @@ PluginLoader.prototype = {
...
+          const acceptedOS = acceptedComponents[AGENT_COMP_ID].os;
+          const acceptedCPU = acceptedComponents[AGENT_COMP_ID].cpu;
+          const setVersions = acceptedComponents[AGENT_COMP_ID].version;
+          const neededEndpoints = acceptedComponents[AGENT_COMP_ID].endpoints;
+
+          if (acceptedOS && acceptedOS.length > 0 && components[AGENT_COMP_ID].os) { // Plugin has manually defined their OS support section
+            if (acceptedOS.includes(components[AGENT_COMP_ID].os)) {
+              console.log(`(${components[AGENT_COMP_ID].os}) is a supported platform for the agent to install (${plugin.identifier}).`);
```
1000TurquoisePogs, line 972: lot of console logs here which should be bootstrapLogger instead.

lib/plugin-loader.js
```diff
@@ -752,71 +756,275 @@ PluginLoader.prototype = {
...
-    for (const pluginDef of sortedAndRejectedPlugins.plugins) { 
-      try {
-        if (this.pluginMap[pluginDef.identifier]) {
-          bootstrapLogger.warn(`ZWED0034W`, plugin.identifier); //bootstrapLogger.warn(`Skipping install of plugin due to existing plugin with same id=${plugin.identifier}`);
+  checkPluginRequirements(plugin, components) {
+    if (plugin.pluginReqs) {
+      const acceptedComponents = plugin.pluginReqs.components;
+      
+      if (acceptedComponents) {
+        if (acceptedComponents[APP_SERVER_COMP_ID]) { /* --- App server component --- */
```
1000TurquoisePogs, line 936: honestly this `checkPluginRequirements` function has no need for explicit component ID checking. Any such need happened in the gathering of component info stage, this can just run through some `Object.keys` instead.

lib/plugin-loader.js
```diff
@@ -48,6 +48,10 @@ const defaultOptions = {
   relativePathResolver: zluxUtil.normalizePath
 }
 
+const APP_SERVER_COMP_ID = 'org.zowe.core.zlux.zlux-core';
```
1000TurquoisePogs, line 51: would rather `app-server` for now.

lib/plugin-loader.js
```diff
@@ -48,6 +48,10 @@ const defaultOptions = {
...
+    
+const AGENT_COMP_ID = 'org.zowe.core.zss';
```
1000TurquoisePogs, line 53: `zss` for now

lib/plugin-loader.js
```diff
@@ -752,71 +756,275 @@ PluginLoader.prototype = {
     let newPlugins = [];
     let successCount = 0;
     const depgraph = new DependencyGraph(this.plugins);
-    for (const pluginDef of pluginDefs) {
-      if (pluginDef.error) {
-        //downstream will pick up error
-        newPlugins.push(pluginDef);
-      } else {
-        depgraph.addPlugin(pluginDef);
+    const componentCapabilities = await this.getComponentCapabilities(pluginContext.config); // Gets the environment and server information
```
1000TurquoisePogs, line 759: potentially this could be cached in the file for reuse, because i think (not certain) that if you do an app install when the server is already running, it will run through this again, so you get call at startup, then 1 call per each app added at runtime. could make it just 1 ever by storing the result.

lib/plugin-loader.js
```diff
@@ -752,71 +756,275 @@ PluginLoader.prototype = {
...
+          }
+        });
+        }).on('error', (e) => {
+          bootstrapLogger.severe(e.message);
+          resolve(); // We don't want to reject here. Error gets caught down stream
+        });
+      }).then(() => {
+        let obj = {};
+        obj[AGENT_COMP_ID] = agentComp;
+        obj[APP_SERVER_COMP_ID] = appServerComp;
```
1000TurquoisePogs, line 924: this is fine but we need some comment in this function to state the purpose:

before checking if dependencies are met, we must learn about the components that exist.
doing this is not formalized currently, so we currently have a block per component to learn about their capabilities, version, environment, etc. perhaps in the future zowe components could have metadata files and/or expected URLs for querying.

lib/index.js
```diff
@@ -299,7 +299,7 @@ Server.prototype = {
         });
       }
     }, installLogger));
-    this.pluginLoader.loadPlugins();
+    yield this.pluginLoader.loadPlugins();
```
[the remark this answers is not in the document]
DivergentEuropeans: Thanks good comment. I believe it works because it's a legacy generator function, basically adding it with an async method that has a promise chain and eventually gets to a then() closes the promise chain, which is basically like closing the last then() of a yield, which is why code continues after the yield and it works
We can see this again with something like, https://github.com/zowe/zlux-server-framework/pull/243/files/1dd87148bb1f85271ad54beb2c51b73adfdb3eed#diff-6d186b954a58d5bb740f73d84fe39073R334 that points to another method, that's not a generator but houses a promise chain

lib/plugin-loader.js
DivergentEuropeans: This is a good comment too and intentional. My reasoning here was if there's an error with the plugin before we even get to the installation step, there is no point in going through the steps of checking for plugin requirement problems. Not to mention, if there happened to be a plugin requirement problem, the pluginDef.error would get replaced by this error and the first, more severe error would get overwritten. Let me add a comment for this, this looks very sloppy without the context

if (pluginDef.error) {
          //downstream will pick up error
          newPlugins.push(pluginDef);
        } else {
          this.checkPluginRequirements(pluginDef, componentCapabilities); // Checks that the environment and server satisfies the plugin req's
          if (pluginDef.error) {
            newPlugins.push(pluginDef);
          } else {
            depgraph.addPlugin(pluginDef);
          }
        }

If we find an error un-related to

lib/plugin-loader.js
DivergentEuropeans: "someone was using trivial auth, in which case this endpoint/functionality would be useless" Yes, useless for the agent environment, but not useless for the app server environment that uses Node commands instead of auth. At least in theory, it could be such that app server has a plugin that doesn't use the mainframe but only runs on a certain OS, maybe like a web Zowe game that uses a specific type of web framework (i.e. MacOS can't do Flash etc) I think I'll keep this one for better safe than sorry, agree that it looks messy though

lib/webapp.js
DivergentEuropeans: The error for bad checks in installPlugins get caught down stream. Success of installPlugins itself doesn't have a return condition, I just made it this way to make it async

lib/assets/i18n/log/messages_en.json
```diff
@@ -292,7 +295,12 @@
   "ZWED0170W":"Plugin (%s) loading failed. Message: \"%s\"",
   "ZWED0171W":"Rejected undefined referrer for url=%s, ip=%s",
   "ZWED0172W":"Rejected bad referrer=%s for url=%s, ip=%s",
+<<<<<<< HEAD
```
lchudinov, line 298: Resolve this merge conflict please.

lib/plugin-loader.js
```diff
@@ -752,71 +782,240 @@ PluginLoader.prototype = {
...
-          successCount++;
-        } else {
-          bootstrapLogger.info("ZWED0125I", pluginDef.identifier); //bootstrapLogger.log(bootstrapLogger.INFO,
-            //`Plugin ${pluginDef.identifier} not loaded`);
-          this.pluginMap[pluginDef.identifier] = {}; //mark it as having failed so it will not be retried
+              if (environmentComponent.os && requiredOS && requiredOS.length > 0 && environmentComponent.os) { // Plugin has manually required their OS support
+                if (requiredOS.includes(environmentComponent.os)) {
+                  bootstrapLogger.debug("ZWED0296I", environmentComponent.os, name, plugin.identifier); //(`(${environmentComponent.os}) is a supported platform for ${name} to install (${plugin.identifier}).`);
+                } 
+                else if (acceptedOS.includes("!"+environmentComponent.os)) {
```
lchudinov, line 983: Where `acceptedOS` was initialized?

lchudinov: Maybe `requiredOS` should be here?

lib/plugin-loader.js
```diff
@@ -752,71 +782,240 @@ PluginLoader.prototype = {
...
+        let rawData = '';
+        res.on('data', (chunk) => { rawData += chunk; });
+        res.on('end', () => {
+          try {
+            const parsedData = JSON.parse(rawData);
+            if (parsedData.agentVersion) {
+              agentComp.version = parsedData.agentVersion;
+            }
+            if (parsedData.arch) {
+              agentComp.cpu = parsedData.arch;
```
lchudinov, line 899: Shouldn't it go to `agentComp.arch`?

lchudinov (changes requested):

lib/plugin-loader.js
DivergentEuropeans: no, arch is meant to be cpu architecture

lib/plugin-loader.js
DivergentEuropeans: Yep

lchudinov (approved):
I checked the PR. Looks good to me. I can see ZWED0174W message if I put wrong `os` into the requirements.

PR #244, opened 2020-08-27
PKCS12 support

timgerstel:
This pull request implements the following feature:
- Add support for certificates and certificate authorities stored within PKCS12 files.

Signed-off-by: Timothy Gerstel <tgerstel@rocketsoftware.com>

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -165,8 +165,10 @@ module.exports.readFilesToArray = function(fileList) {
           p12 = forge.pkcs12.pkcs12FromAsn1(p12Asn1, process.env.KEYSTORE_PASSWORD || "");
         }
         let certData = p12.getBags({bagType: forge.pki.oids.certBag});
-        let certBag = certData[forge.pki.oids.certBag][0];
-        contentArray.push(Buffer.from(forge.pki.certificateToPem(certBag.cert), 'utf8'));
+        for(let i = 0; i < certData.length; i++){
+          let currentCert = certData[forge.pki.oids.certBag][i];
+          contentArray.push(Buffer.from(forge.pki.certificateToPem(currentCert.cert), 'utf8'));
+        }
       } else {
         contentArray.push(content);
       }
--- a/lib/util.js
+++ b/lib/util.js
@@ -148,7 +148,7 @@ module.exports.getOrInit = function(obj, key, dflt) {
   return value;
 };
 
-module.exports.readFilesToArray = function(fileList) {
+module.exports.readFilesToArray = function(fileList, type) {
   var contentArray = [];
   fileList.forEach(function(filePath) {
     try {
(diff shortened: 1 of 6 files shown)


lib/util.js
```diff
@@ -151,7 +152,24 @@ module.exports.readFilesToArray = function(fileList) {
   var contentArray = [];
   fileList.forEach(function(filePath) {
     try {
-      contentArray.push(fs.readFileSync(filePath));
+      let extension = filePath.split('.').pop();
+      let content = fs.readFileSync(filePath);
+      if(extension == 'p12'){
```
lchudinov, line 157: `pfx` extension is a synonym for `p12`. Should the condition take into account both extensions?

timgerstel: Good catch!

lib/util.js
```diff
@@ -151,7 +152,24 @@ module.exports.readFilesToArray = function(fileList) {
...
+        let p12Asn1 = forge.asn1.fromDer(p12Der);
+        let p12;
+        try{
+          p12 = forge.pkcs12.pkcs12FromAsn1(p12Asn1, "");
+        } catch (e) {
+          loggers.bootstrapLogger.warn("ZWED0173W", e.message);
+          p12 = forge.pkcs12.pkcs12FromAsn1(p12Asn1, process.env.KEYSTORE_PASSWORD || "");
+        }
+        let certData = p12.getBags({bagType: forge.pki.oids.certBag});
+        let certBag = certData[forge.pki.oids.certBag][0];
```
lchudinov, line 168: Why should we read only the first certificate in the chain and ignore the remaining?

lchudinov:
@1000TurquoisePogs app-server already supports the `pfx` format. Please clarify, does this PR introduce new functionality?

cleonty (approved):
Looks good to me now.

lchudinov (approved):
Looks good to me now

lib/util.js
```diff
@@ -151,7 +152,26 @@ module.exports.readFilesToArray = function(fileList) {
...
+        let p12Der = forge.util.decode64(content.toString('base64'));
+        let p12Asn1 = forge.asn1.fromDer(p12Der);
+        let p12;
+        try{
+          p12 = forge.pkcs12.pkcs12FromAsn1(p12Asn1, process.env.KEYSTORE_PASSWORD || "");
+        } catch (e) {
+          loggers.bootstrapLogger.warn("ZWED0173W", e.message);
+          p12 = forge.pkcs12.pkcs12FromAsn1(p12Asn1, "");
+        }
+        let certData = p12.getBags({bagType: forge.pki.oids.certBag});
```
1000TurquoisePogs, line 167: p12 files can be like an archive of multiple types of security objects (keys, certificates, crls etc)
the function this is in is generic: the output is assigned to either keys, certs, cert authorities, depending on which section of the processing we are in.

but, this new code appears to assume its only being used for certs, which is not true.

lib/util.js
```diff
@@ -151,7 +152,26 @@ module.exports.readFilesToArray = function(fileList) {
...
+        for(let i = 0; i < certData.length; i++){
+          let currentCert = certData[forge.pki.oids.certBag][i];
+          contentArray.push(Buffer.from(forge.pki.certificateToPem(currentCert.cert), 'utf8'));
```
1000TurquoisePogs, line 170: cert is fine when its a cert, but what if its a key? and, how are certificate authorities different from plain certs? same thing? i'm not 100% sure on that.

PR #245, opened 2020-09-02
Removing usage of async/await for Node 6 support

DivergentEuropeans:
Signed-off-by: Leanid Astrakou <lastrakou@rocketsoftware.com>

### DIFF
--- a/lib/plugin-loader.js
+++ b/lib/plugin-loader.js
@@ -754,10 +754,10 @@ PluginLoader.prototype = {
     });
   },
   
-  async loadPlugins() {
+  loadPlugins: Promise.coroutine(function*() {
     const defs = this.readPluginDefs();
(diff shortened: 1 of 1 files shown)


lib/plugin-loader.js
```diff
@@ -858,10 +858,10 @@ PluginLoader.prototype = {
           count: newPlugins.length
         });
       }
-  },
+  }),
 
   // Note - Not to be confused with auth capabilities, that describe what an auth plugin can do
-  async getComponentCapabilities(config) {
+  getComponentCapabilities: Promise.coroutine(function*(config) {
```
lchudinov, line 864: Why it is a `coroutine` if it returns a Promise? A coroutine can `yield` a promise.

lchudinov (changes requested):

PR #247, opened 2020-09-18
Inject root services into all proxied root services that require auth

lchudinov:
When RBAC enabled there is a need to call `saf-auth` root service from auth handlers. This PR injects root services into all proxied root services that require auth just before the auth middleware.


1000TurquoisePogs (approved):
I don't see a problem with this and I believe it fixes a bug about doing an RBAC check when trying to resolve a proxy service's access rights.

timgerstel (approved):

struga0258 (approved):

PR #249, opened 2020-09-25
Node6 fix: Switch from async to promise and coroutine

1000TurquoisePogs:
A slightly cleaned up version of https://github.com/zowe/zlux-server-framework/pull/245 which uses a promise that is yielded rather than a coroutine.

### DIFF
--- a/lib/plugin-loader.js
+++ b/lib/plugin-loader.js
@@ -754,10 +754,10 @@ PluginLoader.prototype = {
     });
   },
   
-  async loadPlugins() {
+  loadPlugins: Promise.coroutine(function*() {
     const defs = this.readPluginDefs();
-    await this.installPlugins(defs);
-  },
+    yield this.installPlugins(defs);
+  }),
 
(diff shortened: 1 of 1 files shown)


lib/plugin-loader.js
```diff
@@ -858,107 +858,111 @@ PluginLoader.prototype = {
...
+      appServerComp.os = process.platform; // Operating system
+      appServerComp.cpu = process.arch; // CPU architecture
 
-    if (config && config.agent.host) {
-      host = config.agent.host;
-      if (config.agent.http) {
-        port = config.agent.http.port;
-      } else if (config.agent.https) {
-        port = config.agent.https.port;
+      if (config && config.agent.host) {
```
lchudinov, line 874: The `agent` section is optional in `server.json`. I would check `config && config.agent && config.agent.host`.

lchudinov (approved):
It works and looks good to me.

PR #257, opened 2020-11-20
add pluginId/iframe route

NakulManchanda:
Signed-off-by: Nakul Manchanda <nakul.manchanda@ibm.com>
Add support for new destination property to iframe pluginDefinition.json
Add new double iframe default template

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -335,18 +335,10 @@ module.exports.normalizePath = function(oldPath, relativeTo) {
   return normalized;
 }
 
-module.exports.getRemoteIframeTemplate = function() {
-  let defaultRemoteAppTemplate = fs.readFileSync(path.join(__dirname,'default-remote-app-template.html'),"utf-8");
-  return (templatePath) => {
-    let remoteAppTemplate = defaultRemoteAppTemplate;
-    if(templatePath) {
-      remoteAppTemplate = fs.readFileSync(templatePath, "utf-8");
-    }
-    return (remoteUrl) => {
-      return remoteAppTemplate.replace('${remoteUrl}', remoteUrl);
-    }
-  }
-}();
+const defaultRemoteAppTemplate = fs.readFileSync(path.join(__dirname,'default-remote-app-template.html'),"utf-8");
+module.exports.getRemoteIframeTemplate = function(remoteUrl) {
+  return defaultRemoteAppTemplate.replace('${remoteUrl}', remoteUrl);
+}
 
 module.exports.makeRemoteUrl = function(destination,req) {
   return destination
--- a/lib/util.js
+++ b/lib/util.js
@@ -350,7 +350,7 @@ module.exports.getRemoteIframeTemplate = function() {
 
 module.exports.makeRemoteUrl = function(destination,req) {
   return destination
-          .replace('${ZOWE_EXTERNAL_HOST}', req.host)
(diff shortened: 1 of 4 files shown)


NakulManchanda:
@DivergentEuropeans @1000TurquoisePogs this ready for review

lib/util.js
```diff
@@ -335,6 +335,25 @@ module.exports.normalizePath = function(oldPath, relativeTo) {
   return normalized;
 }
 
+module.exports.getRemoteIframeTemplate = function() {
+  let defaultRemoteAppTemplate = fs.readFileSync(path.join(__dirname,'default-remote-app-template.html'),"utf-8");
```
1000TurquoisePogs, line 339: this should be put as a const outside the function call, since you only need to load it once in order to use it multiple times, right?

lib/util.js
```diff
@@ -335,6 +335,25 @@ module.exports.normalizePath = function(oldPath, relativeTo) {
...
+  return (templatePath) => {
+    let remoteAppTemplate = defaultRemoteAppTemplate;
+    if(templatePath) {
+      remoteAppTemplate = fs.readFileSync(templatePath, "utf-8");
```
1000TurquoisePogs, line 343: this must be changed to a promise which uses `readFile` instead of `readFileSync` since this is happening during a network call, it freezes the server while loading the file.
For performance, you may also consider putting the result into a map such that you can get from the map for every time after the first time.
Doing both will greatly improve performance when there are many clients.

lib/webapp.js
```diff
@@ -1798,7 +1798,29 @@ WebApp.prototype = {
...
+      iframeRouterWithMiddleware.push((req,res,next) => {
+          installLog.debug(`Loading iframe app ${plugin.identifier} located at ${plugin.webContent.destination}`);
+          next();
+      });
+      const destination = plugin.webContent.destination;
+      const router = express.Router();
+      router.use((req, res, callback) => {
+        const remoteUrl = zluxUtil.makeRemoteUrl(destination, req);
+        installLog.info(`Loading remote iframe app ${plugin.identifier} located at ${remoteUrl}`);
+        const startingPage = zluxUtil.getRemoteIframeTemplate()(remoteUrl);
```
1000TurquoisePogs, line 1816: alternatively to a map mentioned in previous file, this _installIframePlugin function could hold the template ahead of time, so router.use already has the result.

1000TurquoisePogs (changes requested):
Logically should work, but there are some performance fixes needed.

lib/webapp.js
NakulManchanda: cool, let me try

lib/util.js
NakulManchanda: even before it was a IFFE, but just moved it outside

lib/util.js
NakulManchanda: removed this logic, as not testing this path right now, so just implementing default case for now. Thanks!

lib/webapp.js
NakulManchanda: it just get replace value, after we have remoteURL variable replaced. 
its part of utils module closure

NakulManchanda:
ZLUX Build - with this PR
https://wash.zowe.org:8443/view/Zlux/job/zlux-core/3729/parameters/

1000TurquoisePogs (approved):
Looks good to me now.

PR #263, opened 2021-01-19
APIML Storage

lchudinov:
## APIML Storage
Depends on https://github.com/zowe/zlux-app-server/pull/157

### How to run tests
```shell
  npm install && npm run build
  export GATEWAY_HOST=<gateway-host>
  export GATEWAY_PORT=<gateway-port>
  export CLIENT_KEY=/path/to/keystore/localhost/localhost.keystore.key
  export CLIENT_CER=/path/to/keystore/localhost/localhost.keystore.cer
  npm run test:storage
```

### How to configure Caching Service
1. Add `ZOWE_CACHING_SERVICE_START=true` into `instance.env`
2. Configure VSAM Dataset for Caching Service `ZWE_CACHING_SERVICE_VSAM_DATASET=HLQ.MLQ.CACHE`

Relates to https://github.com/zowe/zlux/issues/446


DivergentEuropeans (approved):
Some minor code clean-up but I agree with the code and schema overall. 🧀 This is nearly identical to our cluster storage APIs. I also tested with the provided apimlStorage test and I think it works well; approving

I have some questions but also a few concerns that aren't relevant to this code but the API in general

DivergentEuropeans:
Please do expand on the test though. I approve but when I was doing the cluster API storage, we missed a few bugs because my test wasn't as complete as it could be...

lib/apimlStorage.ts
```diff
@@ -0,0 +1,438 @@
...
+    }
+    if (typeof response.json !== 'object') {
+      return {};
+    }
+    const json = response.json as { [key: string]: KeyValuePair };
+    const validKeys = Object.keys(json).filter(key => this.keyHasPrefix(key));
+    const result = {};
+    for (const key of validKeys) {
+      const kv = json[key];
+      if (typeof kv.value === 'string') {
```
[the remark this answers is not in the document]
lchudinov: I do serialize/deserialize value into JSON when I store/get it in/from the Cashing Service.

lib/apimlStorage.ts
```diff
@@ -0,0 +1,438 @@
...
+        throw e;
+      }
+    }
+  }
+
+  async deleteAll(): Promise<void> {
+    const data = await this.getAll();
+    for (const key in data) {
+      try {
+        await this.delete(key);
```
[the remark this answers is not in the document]
lchudinov: Yes, deleteAll and getAll are very inefficient for now. As soon as the issue https://github.com/zowe/api-layer/issues/1253 has resolved they will work better.

lib/apimlStorage.ts
```diff
@@ -0,0 +1,438 @@
...
+    const data = await storage.getAll();
+    console.log(`storage contains:`);
+    console.log(`${JSON.stringify(data, null, 2)}`);
+    const newValue = await storage.get(key);
+    console.log(`got new value '${newValue}'`);
+    if (newValue !== value) {
+      throw new Error(`value doesn't match`);
+    }
+    await storage.deleteAll();
+    await storage.delete(key);
```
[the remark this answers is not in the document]
lchudinov: It tests delete by key using empty storage.

lib/apimlStorage.ts
```diff
@@ -0,0 +1,438 @@
...
+      } else {
+        throw e;
+      }
+    }
+  }
+
+  async setAll(dict: { [key: string]: any }): Promise<void> {
+    await this.deleteAll();
+    for (const key in dict) {
+      await this.set(key, dict[key]);
```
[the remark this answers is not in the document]
lchudinov: Again https://github.com/zowe/api-layer/issues/1253

lib/index.js
```diff
@@ -334,6 +340,38 @@ Server.prototype = {
       yield this.apiml.registerMainServerInstance();
     }
   }),
+  
+  configureApimlStorage(apimlConfig) {
+    const { hostname, gatewayPort } = apimlConfig.server;
+    // TODO: This is a temporary way to obtain client cert/key,
+    // it's going to be fixed as soon get more info
+    const clientCertFile = process.env['ZWED_CLIENT_CERT_FILE'];
+    const clientKeyFile = process.env['ZWED_CLIENT_KEY_FILE'];
```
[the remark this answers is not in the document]
lchudinov: We need a special key/cert pair for caching service.

1000TurquoisePogs:
Is it required that the key&cert we use to contact the caching server is NOT the one we use for https? I thought it would be the same. Is it not possible or is having 2 different certs just better somehow?

lib/apimlStorage.ts
```diff
@@ -0,0 +1,382 @@
...
+  'APIML_STORAGE_RESPONSE_ERROR' |
+  'APIML_STORAGE_NOT_CONFIGURED';
+
+export class ApimlStorageError extends Error {
+  constructor(
+    public readonly code: ApimlStorageErrorCode,
+    private cause?: Error,
+    private apimlResponse?: ApimlResponse) {
+    super(code);
+    Object.setPrototypeOf(this, ApimlStorageError.prototype);
```
1000TurquoisePogs, line 77: Is this needed? I thought typescript does this for you.

lib/apimlStorage.ts
```diff
@@ -0,0 +1,382 @@
...
+    return false;
+  }
+  if (!Array.isArray(obj.messages)) {
+    return false;
+  }
+  return true;
+}
+
+function getApimlErrorMessageString(message: ApimlErrorMessage): string {
+  return `${message.messageType}(${message.messageKey}) ${message.messageNumber} ${message.messageContent}`;
```
1000TurquoisePogs, line 134: When the message is printed in the zlux logger, there will be two message types and it could be confusing.
If someone did a log.info on a message type error, it would say
timestamp (zwed:pid) user INFO - ERROR
The message key being inside () is also not the existing logger format.
Maybe this file should just call the zlux logger directly instead of returning toString() results.

lib/apimlStorage.ts
```diff
@@ -0,0 +1,382 @@
...
+  private keyPrefix: string;
+
+  constructor(
+    private pluginId: string
+  ) {
+    this.keyPrefix = this.makeKeyPrefix(this.pluginId);
+  }
+
+  private makeKeyPrefix(pluginId: string): string {
+    return pluginId.replace(/\./g, '');
```
1000TurquoisePogs, line 224: this will cause collisions when products have poorly organized names.
needs a code comment linking to https://github.com/zowe/api-layer/issues/1300

lib/apimlStorage.ts
```diff
@@ -0,0 +1,382 @@
...
+  port: number;
+  tlsOptions: https.AgentOptions;
+}
+
+
+export function makeStorageForPlugin(pluginId: string): ApimlStorage {
+  return new ApimlStorage(pluginId);
+}
+
+const CACHING_SERVICE_URI = '/cachingservice/api/v1/cache';
```
1000TurquoisePogs, line 40: everything on apiml that i have used follows the pattern
/api/v1
/ui/v1/
/ws/v1
Does caching service really start with `/cachingservice`? Thats interesting. Needs a code comment about that.

lib/index.js
```diff
@@ -337,6 +343,42 @@ Server.prototype = {
...
+    if (typeof cachingService !== 'object') {
+      return;
+    }
+    const {cert: clientCertFile, key: clientKeyFile} = cachingService;
+    if (!clientCertFile || !clientKeyFile) {
+      return;
+    }
+    let cert, key;
+    try {
+      cert = fs.readFileSync(clientCertFile);
```
1000TurquoisePogs, line 362: Is it guaranteed to be a file? Is it guaranteed to be a base64 PEM file? Because p12, DER encoding, and keyrings are all possible alternatives for https keys in zowe and I think this code should really be using the logic in webserver.js in order to handle different types.

lib/index.js
```diff
@@ -337,6 +343,42 @@ Server.prototype = {
...
+      return;
+    }
+    if (key && cert && gatewayHost && gatewayPort) {
+      const settings = {
+        host: gatewayHost,
+        port: gatewayPort,
+        tlsOptions: {
+          cert: cert,
+          key: key,
+          rejectUnauthorized: false,
```
1000TurquoisePogs, line 375: can it ever be self signed? shouldnt this just use the tlsOptions from webserver.js?

package.json
```diff
@@ -24,15 +24,22 @@
...
   "scripts": {
-    "build": "tsc"
+    "build": "tsc",
+    "start": "tsc --watch",
+    "test:storage": "mocha --timeout 10000 ./test/storage/*.js"
   },
   "dependencies": {
     "@rocketsoftware/express-ws": "^5.0.0",
     "accept-language-parser": "~1.5.0",
+    "axios": "^0.21.1",
```
1000TurquoisePogs, line 42: To prevent version accidental version updates that change behavior, we use `~`.

1000TurquoisePogs (commented):
I don't think this can be merged until https://github.com/zowe/api-layer/issues/1300 is done, but take a look at my comments meanwhile.

lib/apimlStorage.ts
lchudinov: Yes, it is needed because of https://github.com/Microsoft/TypeScript-wiki/blob/master/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work

lib/apimlStorage.ts
lchudinov: I removed the `messageType` from the message

lib/apimlStorage.ts
lchudinov: Yes, it really does.

package.json
lchudinov: I fixed the version string.

lib/apimlStorage.ts
lchudinov: I added a comment

lib/index.js
lchudinov: Currently, `node.https` are not good to use with Caching Service so we need a way to provide a valid cert/key pair.
I a little bit changed the code: now tlsOptions from `node.https` are used for Caching Service unless they are overridden by `mediationLayer.cachingService.https`. 
Reusing `readTlsOptionsFromConfig` from webserver.js is a good idea. Thanks!

lib/index.js
lchudinov: yes, you are right. see my previous comment. I fixed it.

package.json
```diff
@@ -24,15 +24,22 @@
     {
       "name": "Tim Gerstel",
       "email": "tgerstel@rocketsoftware.com"
+    },
+    {
+      "name": "Leonty Chudinov",
+      "email": "lchudinov@rocketsoftware.com"
```
DivergentEuropeans, line 30: Well deserved

1000TurquoisePogs (approved):

1000TurquoisePogs (approved):

PR #267, opened 2021-02-17
[WIP] Use 'spdy' library for HTTP2

Kreynar:
For some reason it takes a few minutes to load IZSME. 
And it feels like most requests are served slower than previously using HTTP 1.1, but that might be just my imagination, need to do better testing on this.

On webserver.js file you can see, that createServer(...) is called with options object, that contains only 'cert' and 'key' properties. For some reason, if all the old properties are left in options object, then Zowe website doesn't open.


Kreynar:
This PR contained a lot of noise, I'll open a new one

PR #269, opened 2021-02-22
Add loopback ip to referrer check if it is not bound to app server

timgerstel:
This pull request implements the following feature:
- Add loopback ip to referrer check if it exists but is not set to bind to the app servers IP array

Related: https://github.com/zowe/zlux-app-server/pull/143

Signed-off-by: Timothy Gerstel <tgerstel@rocketsoftware.com>

### DIFF
--- a/lib/webapp.js
+++ b/lib/webapp.js
@@ -1263,6 +1263,11 @@ WebApp.prototype = {
           }
         });
       }
+      if(process.env.BIND_TO_LOOPBACK &&
+          (process.env.BIND_TO_LOOPBACK == false || process.env.BIND_TO_LOOPBACK == "false") &&
+          process.env.ZOWE_LOOPBACK_IP){
+        validDomains.push(process.env.ZOWE_LOOPBACK_IP.toLowerCase())
+      }
       if (options.http && options.http.ipAddresses) {
         options.http.ipAddresses.forEach(addr => {
           if (validDomains.indexOf(addr) == -1) {
--- a/lib/webapp.js
+++ b/lib/webapp.js
@@ -1168,13 +1168,13 @@ function makeLoopbackConfig(nodeConfig) {
     return {
       port: nodeConfig.https.port,
       isHttps: true,
-      host: zluxUtil.getLoopbackAddress(nodeConfig.https.ipAddresses)
+      host: nodeConfig.loopbackAddress ? nodeConfig.loopbackAddress : zluxUtil.getLoopbackAddress(nodeConfig.https.ipAddresses)
     }
   } else {
     return {
       port: nodeConfig.http.port,
       isHttps: false,
-      host: zluxUtil.getLoopbackAddress(nodeConfig.http.ipAddresses)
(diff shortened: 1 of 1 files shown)


1000TurquoisePogs (changes requested):
This will not fix the error, but I have a suggestion.
Real error:
Code hits line 1296 `return res.status(403).send('Forbidden due to referrer');` 
It happens because the loopback request has no referrer header, and then the code did not recognize the request as coming from loopback because the IP did not match `this.loopbackConfig.host`

loopbackConfig.host gets set here
https://github.com/zowe/zlux-server-framework/blob/staging/lib/util.js#L309

There are a few problems to fix
* use of https://github.com/zowe/zlux-server-framework/blob/staging/lib/util.js#L284 becomes odd because the purpose of it was to determine which ip is loopback from a list of IPs given. But, the caller now knows what the loopback really is, if it is explicitly specified.

* one way or another, this.loopbackConfig.host must be config.node.loopbackIp if specified.
Two ways come to mind
1) change getLoopbackAddress(listenerAddresses) to getLoopbackAddress(listenerAddresses, nodeConfig) which will allow it to return the right loopback ip given that it can now read the full config
2) dont change it, but conditionally call it depending on whether or not loopbackIp was explicitly specified

* dont read env vars directly in this code. always read config such as config.node.loopbackIp ... env gets translated to config in the other PR https://github.com/zowe/zlux-app-server/pull/143

* ensure that specification of loopbackIp does not necessarily mean to bind to it, as in https://github.com/zowe/zlux-app-server/pull/143

timgerstel:
>     * dont read env vars directly in this code. always read config such as config.node.loopbackIp ... env gets translated to config in the other PR [zowe/zlux-app-server#143](https://github.com/zowe/zlux-app-server/pull/143)
> 

But env vars are used throughout the rest of the code block? https://github.com/zowe/zlux-server-framework/pull/269/files#diff-eaecb33c0e784a4fb1b89c027e9b17fcdcd349e386abcbb4584f187927794598R1249

The other PR assigns the loopback IP to the ipAddresses array which would mean if that order ever changes it would affect this functionality, assuming the best way to distinguish ZOWE_LOOPBACK_ADDRESS in `ZWED_node_https_ipAddresses="${ZOWE_LOOPBACK_ADDRESS},${ZOWE_IP_ADDRESS}",` would just be to reference the first element of the array.

1000TurquoisePogs:
The use of env vars in that code block is also incorrect. Things should stem from the config object so that there is no confusion about where config is coming from. Just 1 place. On the outside, several (env, cli, json file) are mapped to 1 config object, so its the job of convert-env.sh to handle that complexity so that this code can remain simple and flexible.

The other PR assigns the loopback IP to the array CONDITIONALLY, and the particular error we saw was to be fixed on the condition that the loopback IP actually is not added to the array. So, 2 different cases here and we're solving both.

1000TurquoisePogs (approved):

PR #272, opened 2021-03-17
Forward server response up to browser

1000TurquoisePogs:
This will allow users in the desktop to see more about why a login failed, as currently we hide the server message. With this PR, servers are responsible for how much info they disclose as it can be seen in the desktop, but this should have been their responsibility from the beginning, and I don't see any info coming out of zss or apiml that are security risks to forward.


1000TurquoisePogs:
Merging this after testing.

DivergentEuropeans (commented):
I tested this as working with the ZSS handler. Haven't tested API ML handler

PR #274, opened 2021-03-24
Adding code to conditionally contact zss thru apiml

1000TurquoisePogs:
This code cleans up a lot of code related to how to contact zss, in order to enhance the calling to call through apiml if zss is available through apiml.

This seems to work, but it might need additionally a loop to retry contacting apiml like we do with zss directly, to avoid timing errors. But, we do at elast wait until we register with apiml to try to find zss there.

### DIFF
--- a/lib/apiml.js
+++ b/lib/apiml.js
@@ -11,6 +11,7 @@
 const Promise = require("bluebird");
 const eureka = require('eureka-js-client').Eureka;
 const zluxUtil = require('./util');
+const https = require('https');
 
 const log = zluxUtil.loggers.apiml;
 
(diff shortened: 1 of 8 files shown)


lib/plugin-loader.js
```diff
@@ -866,35 +866,17 @@ PluginLoader.prototype = {
...
-          port: port,
-          path: '/server/agent/environment',
-          method: 'GET'
-        };
-        Object.assign(requestOptions, isHttps ? this.tlsOptions : {});
+        const httpApi = requestOptions.protocol == 'https:' ? https : http;
+        requestOptions.path = '/server/agent/environment';
         return new Promise((resolve, reject) => { /* Obtains and stores environment information from agent */
-          httpApi.request(requestOptions, (res) => {
+          httpApi.get(requestOptions, (res) => {
```
lchudinov, line 879: If you change it to `get` then a call `.end()` is not needed anymore.

lib/plugin-loader.js
```diff
@@ -929,16 +911,9 @@ PluginLoader.prototype = {
...
-            hostname: host,
-            port: port,
-            path: '/server/agent/services',
-            method: 'GET'
-          };
-          Object.assign(requestOptions, isHttps ? this.tlsOptions : {});
+          requestOptions.path = '/server/agent/services';
           return new Promise((resolve, reject) => { 
-            httpApi.request(requestOptions, (res) => {
+            httpApi.get(requestOptions, (res) => {
```
lchudinov, line 916: If you change it to `get` then a call `.end()` is not needed anymore.

lib/apiml.js
```diff
@@ -100,6 +101,46 @@ ApimlConnector.prototype = {
       return this.ipAddr;
     }
   }),
+
+  checkAgent(timeout, serviceName) {
+    return new Promise((resolve, reject) => {
+      const options = {
+        method: 'GET',
```
lchudinov, line 108: It looks like `host` and `port` are missing.

lchudinov: You could use Eureka client's method e.g.  https://github.com/zowe/zlux-server-framework/blob/e96f3e876a0914b30684aecfabd3861bde0058e8/lib/apiml.js#L292

1000TurquoisePogs: i fixed it by using the values on `this` since i didnt want to spend time adjusting & testing the code on the eurekarequest function that i havent used before.

PR #275, opened 2021-03-24
Decrease duplicate and verbose logs

1000TurquoisePogs:
This PR moves some info logs to debug, due to being annoying and never having been useful in our troubleshooting.
Many are duplicate in purpose, or possible to derive from other logs.

And when it comes to duplicate logs, we also have a startup problem in cluster mode.
During init, all workers in a cluster do the same tasks in parallel, resulting in many log messages that say the same thing times the number of workers running.
After init, worker logging stops being duplicated due to requests being dispatched to individual workers, so duplication is only a startup problem.

To prevent duplication, I thought it would be good to reduce server core logger verbosity at startup for all workers except the first. Then when startup is finished, restore the verbosity to the requested level.

Another log I hope to eliminate is the warning we get when trying to connect to apiml when it is not yet loaded. I attempt to eliminate this by substituting the eureka client library logger with ours, and suppress the warning during connection.


1000TurquoisePogs:
James tested this as OK, so will merge it.

PR #294, opened 2021-05-21
Fixed keyring warning from terminal proxy

1000TurquoisePogs:

## Proposed changes

This PR addresses Issue: https://github.com/zowe/zlux/issues/512

144w was never a real issue, just a logging error, but the proper way to fix it seems to be to give plugins the list of tls options that the server loads up, since the plugin was trying to load tls options itself and failing and i didn't want to fix it by duplicating logic. however, plugins should not get the same tls object that the server uses itself, so one is cloned.

## Type of change
- [ ] Refactor the code 

## PR Checklist
- [x] New and existing unit tests pass locally with my changes

## Testing

When using keyrings, error 144w would print. With this change, it will not print because the terminal proxy does not try to load key content on its own anymore. the 144w message has been removed from use.


DivergentEuropeans (approved):
App server works OK & changes make sense

PR #316, opened 2021-07-06
Add nonce csp capability

1000TurquoisePogs:
This PR adds the ability for plugins to request use of a csp nonce and have their primary html be a template in which the nonce value is filled in when sent. This should allow plugins such as the desktop to be able to use a nonce in csp rules. Alternative CSP settings could be added in the future to use nonce in more situations.

### DIFF
--- a/lib/webapp.js
+++ b/lib/webapp.js
@@ -199,6 +199,10 @@ function getPresetHeader(name, preset, req) {
       return `default-src 'self' ${req.hostname};`;
     } else if (preset == 'frame-strict') {
       return `frame-src 'self' ${req.hostname};`;
+    } else if (preset == 'script-nonce') {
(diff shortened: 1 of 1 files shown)


lib/webapp.js
```diff
@@ -199,6 +199,10 @@ function getPresetHeader(name, preset, req) {
       return `default-src 'self' ${req.hostname};`;
     } else if (preset == 'frame-strict') {
       return `frame-src 'self' ${req.hostname};`;
+    } else if (preset == 'script-nonce') {
+      let nonce=crypto.randomBytes(16).toString('hex');
```
lchudinov, line 203: I am wondering whether the approach with a random string would be suitable for HA mode when there are multiple app-servers running.

1000TurquoisePogs: Good point, probably not.

PR #317, opened 2021-07-07
Fix issues discovered by LGTM

lchudinov:
## Proposed changes

This PR addresses issues that I discovered using [lgtm](https://lgtm.com/) code scanner. 

## Type of change
- [x] Chore, repository cleanup, updates the dependencies.

### DIFF
--- a/lib/webapp.js
+++ b/lib/webapp.js
@@ -305,7 +305,7 @@ function getAttrib(object, path){
 
 function setAttrib(object, path, value){
   if(object === undefined || path === undefined || 
-      typeof path === 'array' && path.length === 0 || typeof object !== 'object'){
+      Array.isArray(path) && path.length === 0 || typeof object !== 'object'){
     return undefined
   };
   if(typeof path === 'string'){
--- a/lib/webauth.js
+++ b/lib/webauth.js
@@ -169,7 +169,7 @@ AuthResponse.prototype = {
     let result = false;
     for (const type of Object.keys(this.categories)) {
       const authCategoryResult = this.categories[type];
-      if (!(typeof authCategoryResult) === "object") {
+      if (typeof authCategoryResult !== "object") {
         continue;
       }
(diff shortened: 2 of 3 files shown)


lib/webapp.js
```diff
@@ -305,7 +305,7 @@ function getAttrib(object, path){
 
 function setAttrib(object, path, value){
   if(object === undefined || path === undefined || 
-      typeof path === 'array' && path.length === 0 || typeof object !== 'object'){
+      Array.isArray(path) && path.length === 0 || typeof object !== 'object'){
```
lchudinov, line 308: The result of this 'typeof' expression was compared to 'array', but the two could never be equal.

lib/webauth.js
```diff
@@ -169,7 +169,7 @@ AuthResponse.prototype = {
     let result = false;
     for (const type of Object.keys(this.categories)) {
       const authCategoryResult = this.categories[type];
-      if (!(typeof authCategoryResult) === "object") {
+      if (typeof authCategoryResult !== "object") {
```
lchudinov, line 172: This expression was of type boolean, but it was compared to an expression of type string.

plugins/config/lib/configService.js
```diff
@@ -2328,7 +2328,7 @@ function getJSONFromLocation(relativeLocation,directories,startScope,endScope,pl
       if (updatedConfiguration) {
         logger.debug("ZWED0256I", JSON.stringify(updatedConfiguration)); //logger.debug("Configuration is now = "+JSON.stringify(updatedConfiguration));
         var filesFound = Object.keys(updatedConfiguration);
-        for (var i = 0; i < filesFound; i++) {
+        for (var i = 0; i < filesFound.length; i++) {
```
lchudinov, line 2331: There was missing `.length` in comparison.

1000TurquoisePogs (approved):

PR #323, opened 2021-08-19
Two user-found bugfixes.

1000TurquoisePogs:

## Proposed changes

This fixes 2 problems found by users: 
1) timeout can occur at cluster storage init because something we thought would never take a long time did.
2) we have a regression where the server cannot bind to a hostname. this fixes it.

## Type of change

## PR Checklist
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] New and existing unit tests pass locally with my changes

## Testing

Testing the cluster timeout is difficult because it was only reproduced on a slow system, but testing the hostname fix is easy: set node.http.ipAddresses to a hostname rather than IP, and watch as before the server would throw and exception. With this fix the server will load after it resolves the hostname to ip via dns.

### DIFF
--- a/lib/clusterManager.js
+++ b/lib/clusterManager.js
@@ -680,8 +680,14 @@ if (cluster.isMaster) {
         return resultHandler[0];
       },
       function(e) {
-        clusterLogger.warn("ZWED0167W - Error adding to the storage: " + e);
-      }
+        if (e.startsWith('Timeout call')) {
+          clusterLogger.severe("ZWED0144E - Cluster storage object timeout." +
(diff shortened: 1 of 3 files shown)


lib/clusterManager.js
```diff
@@ -680,8 +680,14 @@ if (cluster.isMaster) {
         return resultHandler[0];
       },
       function(e) {
-        clusterLogger.warn("ZWED0167W - Error adding to the storage: " + e);
-      }
+        if (e.startsWith('Timeout call')) {
```
DivergentEuropeans, line 683: We have three places we reject with 'Timeout call' yet we hardcode "Timeout call " on line 533. If that ever changes, those places would need changes too, this seems like a good spot to put a CONST

DivergentEuropeans (approved):

PR #332, opened 2021-09-10
Enable app-server to reach apiml when in kubernetes if using self-signed

1000TurquoisePogs:
VERIFY_CERTIFICATES=true/false should apply globally but seems not to be the case in app-server. This is a fix for the case of contacting apiml for auth actions.


1000TurquoisePogs:
Merging after testing.

PR #350, opened 2021-12-01
Tracing auth handler server activity

1000TurquoisePogs:
To better understand cases such as https://github.com/zowe/zlux/issues/761
This PR is just to add some more messages so we can learn if there are unexpected calls to servers that would result in multiple auth events rather than 1.

### DIFF
--- a/lib/webauth.js
+++ b/lib/webauth.js
@@ -228,6 +228,7 @@ module.exports = function(authManager, cookiePort, isSecurePort) {
           let authPluginSession = getAuthPluginSession(req, pluginID, {});
           let caps = handler.getCapabilities();
           if (caps.canLogout === true) {
+            authLogger.debug(`${req.session.id}: Initiating logout with ${handler.pluginID}`);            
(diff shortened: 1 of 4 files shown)


1000TurquoisePogs:
To view this trace output, set this object within server.json
```
"logLevels": {
  "_zsf.routing": 5,
  "_zsf.auth": 5,
  "org.zowe.zlux.auth.safsso.*": 5
}
```

lib/webauth.js
```diff
@@ -228,6 +228,7 @@ module.exports = function(authManager, cookiePort, isSecurePort) {
           let authPluginSession = getAuthPluginSession(req, pluginID, {});
           let caps = handler.getCapabilities();
           if (caps.canLogout === true) {
+            authLogger.debug(`${req.session.id}: Initiating logout with ${handler.pluginID}`);
```
DivergentEuropeans, line 231: This one is already being logged with ZWED0205I

DivergentEuropeans (approved):
Overall good tracing change 👍

PR #357, opened 2022-01-05
Do not retry login operations on failure

1000TurquoisePogs:
There are a few cases in the apiml handler code where a different type of login is attempted if the first one fails, or a retry is attempted upon exception. Instead, a failure should give up in order to not cause lockout.

### DIFF
--- a/plugins/sso-auth/lib/apimlHandler.js
+++ b/plugins/sso-auth/lib/apimlHandler.js
@@ -147,20 +147,14 @@ class ApimlHandler {
     if (request.body) {
       this.logger.debug(`Authenticate with body`);
       return new Promise((resolve, reject) => {
-        this.doLogin(request, sessionState).then(result=> {
-          if (result.success === true) {
-            resolve(result);
-          } else {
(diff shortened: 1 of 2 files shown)


plugins/sso-auth/lib/apimlHandler.js
```diff
@@ -147,20 +147,14 @@ class ApimlHandler {
...
+          resolve(result);
         }).catch(e=> {
-          this.authenticateViaCookie(request, sessionState).then(result=> resolve(result))
-            .catch(e => reject(e));
+          Promise.resolve({success: false});
         });
       });
     } else if (request.cookies && request.cookies[TOKEN_NAME]) {
-      return this.authenticateViaCookie(request, sessionState);
+      return this.authenticateViaCookie(request, sessionState, true);
```
lchudinov, line 157: My concern is the third argument because the function `authenticateViaCookie(request, sessionState)` accepts only two arguments. Is this a typo or do you have uncommitted changes for `authenticateViaCookie`?

lchudinov (dismissed):

DivergentEuropeans (approved):

PR #370, opened 2022-01-24
Update packages to what we can do without changes

1000TurquoisePogs:
https://github.com/zowe/zlux-server-framework/pull/369 but for staging


timgerstel (approved):
Builds fine with new package versions

PR #387, opened 2022-02-21
Fixing direct routing - V2

sakshibobade21:
## Proposed changes

PR attempts to make requests to ZSS w/o any loopbacks to the App server.

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactor the code 

## PR Checklist

### DIFF
--- a/lib/webapp.js
+++ b/lib/webapp.js
@@ -987,6 +987,7 @@ WebServiceHandle.prototype = {
           req.method = requestOptions.method;
           req.headers = requestOptions.headers;
           req.rawHeaders=[];
+          req._zsfinternal = true
           let keys = Object.keys(req.headers);
           for (let i = 0; i < keys.length; i++) {
             let k = keys[i];
@@ -1269,7 +1270,7 @@ headers: {
     return function localCheck(req, res, next) {
       const loopbackData = req.get(ZLUX_LOOPBACK_HEADER);
       const address = ipaddr.process(req.ip)
-      if (loopbackSecret == loopbackData && (req.ip == localIp || address.range() == 'loopback')) {
+      if (loopbackSecret == loopbackData && ((req.ip == localIp || address.range() == 'loopback') || req._zsfinternal)) {
         next();
       } else {
         res.status(403).json({"error":"Cannot access requested service externally"});
--- a/lib/webapp.js
+++ b/lib/webapp.js
@@ -934,7 +934,9 @@ WebServiceHandle.prototype = {
         serviceModified = false;
         const reqPlugin = pluginsArray.filter(plugin => plugin.identifier === this.service.sourcePlugin);
         reqPlugin.filter(plugin => {
-          reqDataservices = plugin.dataServices.filter(dataService => dataService.name === this.service.sourceName);
+          if(plugin.dataServices) {
+            reqDataservices = plugin.dataServices.filter(dataService => dataService.name === this.service.sourceName);
+          }
         })
         if(reqDataservices) {
           const sourcePlugin = this.service.sourcePlugin;
@@ -945,8 +947,7 @@ WebServiceHandle.prototype = {
       }
 
       if(this.service && this.service.imported) {
-        const reqPath = requestOptions.path.split('/').slice(6).join('/');
-        requestOptions.path = `/ZLUX/plugins/${this.service.sourcePlugin}/services/${this.service.sourceName}/${reqPath}`
+        requestOptions.path = zLuxUrl.makePluginURL('ZLUX', this.service.sourcePlugin) + zLuxUrl.makeServiceSubURL(this.service, false, false, path);
         routingLog.debug(`Call agent host=%s import resolved to path=%s`,requestOptions.hostname, requestOptions.path);
       }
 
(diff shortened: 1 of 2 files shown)


lib/webapp.js
```diff
@@ -1003,6 +1026,17 @@ WebServiceHandle.prototype = {
       }
       //if not internal routing
       if (httpOrHttps) {
+        // Removing the version from requestOptions.path if request is for zss
+        if(requestOptions.hostname === this.environment.agentRequestOptions.host && requestOptions.port === this.environment.agentRequestOptions.port) {
+          const versionPattern = new RegExp('^(\\d+\\.)?(\\d+\\.)?(\\*|\\d+)$');
+          const reqPath = requestOptions.path;
+          const pathArray = reqPath.split('/');
+          const reqPathVersion = pathArray.length >= 7 ? pathArray[6] : '';
+          if(versionPattern.test(reqPathVersion)) {
```
1000TurquoisePogs: This seems like a hack and needs a code comment with an example to explain why it works because it will probably break later.
app-server URLs are in the shape of
`/ZLUX/plugins/${service.pluginname}/services/${service.name}/${version}/${reqPath}`
zss URLs are in the shape of
`/ZLUX/plugins/${service.pluginname}/services/${service.name}/${reqPath}`
${version}=a period-separated number, or '_current' (which i think is resolved to a number at this point but not certain!)
Examples include 
/ZLUX/plugins/org.zowe.zlux.sample.angular/services/appServerStorage/1.0.1/asd
/ZLUX/plugins/org.zowe.zlux.sample.angular/services/appServerStorage/_current/asd

One part I'm not certain about is what is `(\\*|\\d+)` at the end trying to do with `\\*`?

lib/webapp.js
```diff
@@ -1003,6 +1026,17 @@ WebServiceHandle.prototype = {
...
+            requestOptions.path = reqPath.replace('/'+reqPathVersion, '');
```
1000TurquoisePogs: this should be a replace on pathArray[6] and then join the pathArray, rather than a replace on the entire string since it's possible a different part of the URL before [6] matches.

lib/webapp.js
```diff
@@ -928,6 +929,27 @@ WebServiceHandle.prototype = {
...
+        if(reqDataservices) {
+          const sourcePlugin = this.service.sourcePlugin;
+          const sourceName = this.service.sourceName;
+          this.service = {...reqDataservices[0], sourcePlugin: sourcePlugin, sourceName: sourceName, imported: true };
+          serviceModified = true;
+        }
+      }
+
+      if(this.service && this.service.imported) {
+        const reqPath = requestOptions.path.split('/').slice(6).join('/');
```
1000TurquoisePogs: the argument `path` at the top is already this.

lib/webapp.js
```diff
@@ -928,6 +929,27 @@ WebServiceHandle.prototype = {
...
+        requestOptions.path = `/ZLUX/plugins/${this.service.sourcePlugin}/services/${this.service.sourceName}/${reqPath}`
```
1000TurquoisePogs: don't hardcode this logic as it could change and break later, use utility functions. Try using or expanding the content here https://github.com/zowe/zlux-server-framework/blob/343675ebff4b949b1a775dd9d45f2e5c0f7967c0/lib/url.js#L13-L30

1000TurquoisePogs (dismissed):

lib/webapp.js
```diff
@@ -928,6 +929,27 @@ WebServiceHandle.prototype = {
         requestOptions.headers = headers;
       }
 
+      while(this.service && this.service.type === 'import' && serviceModified) {
+        let reqDataservices;
+        serviceModified = false;
+        const reqPlugin = pluginsArray.filter(plugin => plugin.identifier === this.service.sourcePlugin);
+        reqPlugin.filter(plugin => {
+          reqDataservices = plugin.dataServices.filter(dataService => dataService.name === this.service.sourceName);
```
DivergentEuropeans: Wouldn't this produce a JS error if plugin.dataServices doesn't exist? As I understand it, dataServices are optional to plugins though I'm not all up to date with the context for this change

lib/webapp.js
```diff
@@ -928,6 +929,27 @@ WebServiceHandle.prototype = {
...
+        })
+        if(reqDataservices) {
+          const sourcePlugin = this.service.sourcePlugin;
+          const sourceName = this.service.sourceName;
+          this.service = {...reqDataservices[0], sourcePlugin: sourcePlugin, sourceName: sourceName, imported: true };
```
DivergentEuropeans, line 944: If `this.service.sourcePlugin` exists and `this.service.sourceName` exists, shouldn't this just be:
`this.service = {...reqDataservices[0], this.service, imported: true };` or did you intend on removing the other service properties (`service.type` etc)?

lib/webapp.js
```diff
@@ -2015,6 +2049,7 @@ WebApp.prototype = {
     }
     this._resolveImports(plugin, urlBase);
     this.plugins.push(plugin);
+    pluginsArray.push(plugin);
```
DivergentEuropeans, line 2054: Also I'm not seeing any major difference between `this.plugins` and `pluginsArray` am I missing something? One gets defined at the start of the file, the other gets defined in constructor. They both get pushed the same stuff & they don't get altered later in this file

lib/webapp.js
sakshibobade21: Yes. I have intended on removing the `service.type` property. `service.type` gets updated depending upon the value in `reqDataservice[0]`

lib/webapp.js
sakshibobade21: I have used `pluginsArray` because `this.plugins` is not accessible inside `call()`. Please let me know if I am missing something.

DivergentEuropeans: I see 2 ways:

1. `WebApp `eventually creates `WebServiceHandle` 's as I understand, so `this.plugins` could be passed through any method that creates `WebServiceHandle`, to make maybe `WebServiceHandle.plugins` or
(better, easier?) 
2. Make `this.plugins` a global object for the file, instead of a `WebApp` object

DivergentEuropeans (approved):
I don't know how to test this

However, the code looks good. Changes are addressed. At a glance, I don't see any regressions. If our internal testing is satisfied, we can merge

PR #388, opened 2022-03-03
Fix install app outdated variable

1000TurquoisePogs:
app-server instanceDir predates zowe 1.0 and is most similar to workspacedirectory, so we update missing INSTANCE_DIR in v2 to instead just use instanceDir


1000TurquoisePogs:
Add a note at the top of this file to say this uses environment variables rather than a config file due to the context in which it is run being more than just the normal server startup

timgerstel (approved):

PR #394, opened 2022-03-11
V1 - Updating the routing

sakshibobade21:
## Proposed changes

PR attempts to make requests to ZSS w/o any loopbacks to the App server.

## Type of change
- [ ] Refactor the code 

## PR Checklist

### DIFF
--- a/lib/webapp.js
+++ b/lib/webapp.js
@@ -1001,9 +1001,19 @@ WebServiceHandle.prototype = {
             
           }
 
+          // Populating req.query with the query parameters
+          const queryString = path.includes('?') ? path.substring(path.indexOf('?') + 1) : '';
+          if(queryString !== '') {
+            const queryStringArray = queryString.split('&');
+            queryStringArray.forEach(pair => {
+              pair = pair.split('=');
+              req.query[pair[0]] = pair[1] || '';
+            })
+          }
+
           if (options.body) {
-            req._body = options.body;
-            req.body = options.body;
+            req._body = JSON.parse(options.body);
+            req.body = JSON.parse(options.body);
           } else {
             delete req._body;
             delete req.body;
--- a/lib/webapp.js
+++ b/lib/webapp.js
@@ -1012,11 +1012,11 @@ WebServiceHandle.prototype = {
           }
 
           if (options.body) {
-            try {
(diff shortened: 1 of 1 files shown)


lib/webapp.js
```diff
@@ -1001,9 +1001,19 @@ WebServiceHandle.prototype = {
...
+              pair = pair.split('=');
+              req.query[pair[0]] = pair[1] || '';
+            })
+          }
+
           if (options.body) {
-            req._body = options.body;
-            req.body = options.body;
+            req._body = JSON.parse(options.body);
+            req.body = JSON.parse(options.body);
```
1000TurquoisePogs: This has several problems.
1) if you parse and its not json, you will throw an exception
2) options.body type detection is already done around line 913
3) line 923 does JSON.stringify(), so this just reverses it. Choose one or the other.
4) isn't stringify better since thats what the network does? what issue is this solving?

1000TurquoisePogs (changes requested):

lib/webapp.js
sakshibobade21: Yes, stringify is better.
But Robert's team had created the test plugins to reproduce few errors.
And one of the errors occurred because they were accepting the parsed JSON body.

And hence I used JSON.parse() only for the requests that are routed internally. 
So we still need to do JSON.stringify() at line 923 for other requests.

1000TurquoisePogs: OK, thank you for the explanation. A try-catch is still needed however because the body might not be json! If the body is a string, boolean, number... this will not work for other cases. Please check `headers["Content-Type"]` that was set in 923 or just optimistically try and then catch by saying that well, its not json I guess.

Also a code comment about why you're doing this will be great for the future when we wonder why it is the way it is.

1000TurquoisePogs (changes requested):

lib/webapp.js
sakshibobade21: Thank you.
Please let me know if the changes looks good to you or it's still more safer to use the try catch instead of if else.

1000TurquoisePogs (approved):

1000TurquoisePogs:
Needs DCO fix.

PR #428, opened 2022-04-25
Fix for v2 keyring / dev env

1000TurquoisePogs:


1000TurquoisePogs:
https://github.com/zowe/zowe-install-packaging/actions/runs/2220673217  shows this fix works.

PR #439, opened 2022-07-27
Bugfix for external route bad url

1000TurquoisePogs:
In v1.28 there was a regression where external type service requests would be called with only their prefix, not their full path.
This fixes it, also fixes logging in the call() function, and improves logging there a little too. Also, we better protect against double slashes in the URL


1000TurquoisePogs:
Merging this after getting a review from Robert and testing by his team to confirm it fixed their bug.

PR #440, opened 2022-07-31
Fix for auth when using only zss

1000TurquoisePogs:
app-server was expecting zss to have a cookie suffix of port conditional to ha setup, but zss unconditionally uses zowe.cookieIdentifier now, so app-server must expect the same.

## Type of change

## PR Checklist

## Testing
Turn off gateway, discovery, caching, and api-catalog server and try to login just using zss + app-server.


1000TurquoisePogs:
Looks like zss already returned to previous cookie behavior in v2.3 staging so this PR would break in the opposite way.
Closing this for now.

PR #449, opened 2022-11-23
Redirect to gateway homepage when possible

1000TurquoisePogs:
To avoid cases like https://github.com/zowe/zlux/issues/863 this PR tries to redirect `/` to the desktop homepage on the gateway when possible, rather than continuing to the desktop directly.


DivergentEuropeans (approved):
It works now 👍

PR #462, opened 2023-02-06
Allow CAs to be passed into callService, or default

1000TurquoisePogs:
This PR allows a dataservice that issues a call() to provide CAs to associate with that call, overriding the default.

The default CAs will be sufficient for calls that occur within servers known to zowe's core via keystore configuration. This is generally the servers under the apiml gateway sphere. If your service needs to contact a server elsewhere, and needs its CA for verification, then you can override the CA list on the call() when you provide the call() options object.
The CAs will only be used if certificate verification is enabled, which is determined at the server config level.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This repo is part of the app-server Zowe Component, and the change logs here may
 ## 2.8.0
 
 - Enhancement: Support zowe.verifyCertificates=NONSTRICT (#468)
+- Enhancement: Allow dataservices to pass CAs into the call() function (#462)
 
 ## 2.3.0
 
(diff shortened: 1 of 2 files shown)


sakshibobade21 (approved):

DivergentEuropeans (commented):
We later reset the request's CA with your default case (`this.environment.agentRequestOptions.ca`)
if `!(this.isAgentService || (this.service && (this.service.type === 'external' || this.service.type === 'service')))` anyway down at:
https://github.com/zowe/zlux-server-framework/pull/462/files#diff-eaecb33c0e784a4fb1b89c027e9b17fcdcd349e386abcbb4584f187927794598R971

so just making sure this order of operations/conditionals is how you wanted it to be

1000TurquoisePogs:
that is done because if the destination is agent, we know that the ca which should be used is the agent ca. attempts by the user to choose a different ca would invite user error.

PR #471, opened 2023-05-09
Update apimlHandler.js

hrshkshri:
The catch block in the if (request.body) clause of the authenticate method returns a new Promise that resolves with an object {success: false}. However, this Promise is not returned from the authenticate method, and therefore, the result will not be used or processed by the calling code.


1000TurquoisePogs (approved):

1000TurquoisePogs:
Resolves https://github.com/zowe/zlux/issues/964

PR #493, opened 2023-07-31
Add more no-caching headers

1000TurquoisePogs:
The server currently has a way you can decide if you want your network calls to be cached by the browser or not. Usually, you'd want to cache static content but not dynamic content like an api.
The headers we set for no-caching weren't sufficient for all browsers, so security scanners recommend this extra set of headers in this PR.


struga0258 (approved):

1000TurquoisePogs:
DCO bot was stalled, so I force-merged it.

PR #511, opened 2023-10-11
Use zowe standard network config

1000TurquoisePogs:
This PR uses the schema in https://github.com/zowe/zowe-install-packaging/pull/3509 to allow users to more easily customize network attributes for IP and TLS level.

CHANGELOG: Added support for using "zowe.network" and "components.app-server.zowe.network" to set listener IP and TLS version properties.
VERSION: 2.13.0

How to test: Change the IP to something that doesn't work, and you should see you can't connect :)
For TLS, you can customize what version of TLS firefox listens for, as well as view what version was used when connecting, so you can set different TLS levels and observe them through firefox.


sakshibobade21 (approved):

sakshibobade21:
The PR looks good to me. Approved it.

PR #517, opened 2023-11-22
Update TLS configuration to disallow downgrades

1000TurquoisePogs:
After discussion, it was decided it would be best not to introduce the feature for allowing users to downgrade their TLS setup. We'll keep it at 1.2 minimum unless there is some high demand otherwise.


DivergentEuropeans (approved):
This looks right

PR #521, opened 2023-12-11
Fix for detecting whether or not in HA mode

1000TurquoisePogs:
App-server was determining whether or not to do HA-related actions depending on whether it detected an environment variable that appears to never have been in v2, and maybe was from v1.

Now, I try to detect if we're in an HA mode depending environment variables that fit the pattern of `ZWE_haInstances_<something here>_hostname` but NOT `ZWE_haInstances_<something here>_components_<...>hostname`

This pattern attempts to count the quantity of items in the haInstances object, without counting references to hostname that could instead be present in some components.

Usually we would read the zowe yaml for this, but the existing, exported function doesnt take the config as an argument, so I do not want to break any API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to the Zlux Server Framework package will be documented in this file..
 This repo is part of the app-server Zowe Component, and the change logs here may appear on Zowe.org in that section.
+
+## 2.14.0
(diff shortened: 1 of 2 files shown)


1000TurquoisePogs:
Fix for https://github.com/zowe/zlux/issues/969

timgerstel (approved):

PR #529, opened 2024-01-22
Add zcrypto library use for jceccaracfks support

1000TurquoisePogs:
To address https://github.com/zowe/community/issues/1851 for the app-server, the https://www.npmjs.com/package/zcrypto library uses gsk/systemssl for getting keyrings in node.
ZSS is also a user of gsk, and was tested as working for ICSF, so I am hopeful that switching away from our own keyring_js library towards this library will yield better long-term results for keyring support within the app-server.

In this PR, both libraries are still available, though only switchable via a const that is not yet exposed as a configuration item.

How to test:
In zowe's config, you want to use keyrings of either JCERACFKS type or of JCECCARACFKS.
We already support JCERACFKS, so we should see no regression (server loads and is accessible for a url such as the desktop or /plugins)
If we're successful with JCECARACFKS, we should see the same behavior, whereas before we could not load the key/cert at all.

### DIFF
--- a/lib/webserver.js
+++ b/lib/webserver.js
@@ -29,11 +29,18 @@ const contentLogger = util.loggers.contentLogger;
 const childLogger = util.loggers.childLogger;
 const networkLogger = util.loggers.network;
 
+const USE_KEYRING_JS = false;
+
 const os = require('os');
 let keyring_js;
+let zcrypto;
 try {
   if (os.platform() == 'os390') {
-    keyring_js = require('keyring_js');
+    if (USE_KEYRING_JS) {
+      keyring_js = require('keyring_js');
+    } else {
+      zcrypto = require('zcrypto');
+    }
   }
 } catch (e) {
   bootstrapLogger.warn('Could not load zcrypto library, SAF keyrings will be unavailable');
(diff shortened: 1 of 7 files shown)


lib/webserver.js
```diff
@@ -23,22 +23,28 @@ const util = require('./util');
...
 
+const USE_KEYRING_JS = false;
+
 const os = require('os');
-let keyring_js;
+let KeyringHandler;
 try {
   if (os.platform() == 'os390') {
-    keyring_js = require('keyring_js');
+    if (USE_KEYRING_JS) {
```
1000TurquoisePogs, line 39: turn this into a zowe.yaml config parameter

lib/webserver.js
```diff
@@ -300,7 +321,11 @@ WebServer.prototype = {
             let certificateList;
             if(userId && keyringName) {
               try {
-                certificateList = keyring_js.listKeyring(userId, keyringName);
+                let includingPrivateKey = false;
```
1000TurquoisePogs, line 324: add a note saying this will only work if there are truly only 2 types of objects in a keyring, but thus far that's all i've seen when working with users.

1000TurquoisePogs:
Seems not possible due to PKDS keys not being exportable via this technique. Instead, AT-TLS must be used.

PR #544, opened 2024-04-05
Fix ability to use attls

1000TurquoisePogs:
Fix not being able to run in an http mode and still do https client calls for use with attls

ATTLS used to work fine but we've had regressions about using http.
There are conditionals in the code where if http exists, https will not be used.
There are conditionals in the code that detect if https exists, and to not use http if so.
These are not quite correct and should be based upon attls use.
The server can be HTTP for attls, yet still wish to issue HTTPS calls outward, and therefore want certificates still.

This PR depends upon https://github.com/zowe/zowe-install-packaging/pull/3463 and https://github.com/zowe/zlux-app-server/pull/300


lib/util.js
```diff
@@ -496,13 +497,25 @@ const isHaMode = module.exports.isHaMode = function () {
...
 
-//TODO, ATTLS
 module.exports.isServerHttps = function(zoweConfig) {
-  return !!zoweConfig.components['app-server'].node.https;
+  return Number.isInteger(zoweConfig.components['app-server'].node.https?.port);
 }
 
+function isClientAttls(zoweConfig) {
+  let clientGlobalAttls = zoweConfig.zowe.network?.client?.tls?.attls;
+  let clientLocalAttls zoweConfig.components['app-server'].zowe?.network?.client?.tls?.attls;
```
sakshibobade21: We need "=" after `let clientLocalAttls`

sakshibobade21 (approved):
Just found one small typo. Otherwise the PR lgtm.

PR #565, opened 2024-08-24
Fixing the SAF keyring error through console logging it.

Atharva-Kanherkar:
## Proposed changes
 The SAF keyring error was an error which did not allow the server to start without an error, until we could find the fix, we have decided to bypass it and know the error through console logging, 

## Type of change
- [ ] Refactor the code 

## PR Checklist


1000TurquoisePogs (changes requested):
Disabling an error to work-around a bug is not a solution to a bug.
Identify why the bug has happened and resolve that.

If this PR is based on an outdated copy of v2.x/staging, you may have missed that this was already fixed.

https://github.com/zowe/zlux-server-framework/commit/a7e5b99020ae8ef67b77c7f1cf61fff008928940

There was a bug in which `saf` was changed from being initialized to undefined to instead be initialized to an empty array.

Previously `if (saf)` was done, which was previously false but then true when  set to an empty array.

A fix was done to do `if (saf.length > 0` instead.
This fix was made to v2.x/rc branch but not v2.x/staging branch.
Therefore the bug still existed in v2.x/staging branch.


But, the two branches were synchronized a week or two ago so you should check if the problem that you still have exists. You may not need a PR at all.

PR #572, opened 2024-10-14
Log in local timezone if configured to

1000TurquoisePogs:
This PR checks the zowe property `zowe.logging.timezone: local` via environment variable `ZWE_zowe_logging_timezone` (due to early load of logging, the yaml isnt available)
when this is set to "environment", the server will get the timezone from nodejs (nodejs knows system timezone, seems to respect the TZ environment variable)

It will apply this timezone as an offset so that the server can log in local time instead of UTC time.
The default is still UTC time via `zowe.logging.timezone: utc`


timgerstel (approved):
LGTM

PR #580, opened 2024-12-20
Fix Eureka Registration when ATTLS (v3)

1000TurquoisePogs:
app-server should use "utils.isClientAttls()" to determine if outbound network requests should be https (if not attls) or http (if attls, because then attls adds the tls, becoming https in the end).

It appears that apiml.js does not do these checks when doing eureka registration, so when attls is used, it does double-tls, resulting in the repeated error:

Eureka request failed to endpoint ..., next server retry...

This error can be seen under normal circumstances due to a timing issue, but eventually resolves to a message

" registered with eureka:..."

In the case of attls, due to the double-tls failure, success never occurs and the "eureka request failed to endpoint" continues forever.

In this PR, I use "isClientAttls" in a few places.
In my testing, it is only truly needed in "getServiceUrls()"
However, through reading documentation of https://www.npmjs.com/package/@rocketsoftware/eureka-js-client and checking other areas in this code, I identified 4 places where it would be best to be explicitly either http or https, rather than allowing the eureka-js-client library to do some implicit behavior that may change in the future.

## Type of change

## PR Checklist

## Testing
* Toggle "zowe.network.server.tls.attls" and "zowe.network.client.tls.attls" so that they are either both true or both false.
* When both false, there should be no regression. Its the default and currently works.
* When both true, " registered with eureka:..." should be seen in log. It is seen whenever you have a valid network configuration (AT-TLS or not) between app-server and discovery.
* seeing "Eureka request failed to endpoint ..., next server retry..." in the log at first is normal, but if it continues without " registered with eureka:..." then it is a problem.
* env var "NODE_DEBUG: tls,request" can be used to observe what network activity app-server does before these "eureka" messages appear. You should see "http://" urls and no "TLS" messages when AT-TLS is enabled. You should see "https://" and "TLS" messages when AT-TLS is disabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 All notable changes to the Zlux Server Framework package will be documented in this file.
 This repo is part of the app-server Zowe Component, and the change logs here may appear on Zowe.org in that section.
 
+## 3.1.0
+- Bugfix: App-server could not register with discovery server when AT-TLS was enabled for app-server. (#580)
+
 ## 3.0.0
 - Enhancement: Add ability for server to dynamically load plugin web content based on `entryPoint` specification in the
 `pluginDefinition.json`
--- a/lib/apiml.js
+++ b/lib/apiml.js
@@ -171,7 +171,7 @@ ApimlConnector.prototype = {
       httpEnabled: false,
       httpsEnabled: true
     };
-    const proto = 'https';
+    const proto = this.isClientAttls ? 'http' : 'https';
 
     log.debug("ZWED0141I", proto, this.port); //"Protocol:", proto, "Port", port);
     log.debug("ZWED0142I", JSON.stringify(protocolObject)); //"Protocol Object:", JSON.stringify(protocolObject));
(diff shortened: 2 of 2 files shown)


pablocarle:
Hi, I wonder how does the desktop handle the registration itself under this context?
I'm asking more precisely about the metadata, it would be assumed that if outbound rule is enabled for Zlux -> Discovery, it may be also enabled for Gw -> Zlux, so the question is if zlux will be reached using HTTP in routing.

1000TurquoisePogs:
@pablocarle in communication there are places that must be aware of ATTLS and other places that should not be aware.

The app-server must be aware of ATTLS when sending traffic out, so that it does not create double-TLS. So, it must send as HTTP when ATTLS is on.

But because the receiver (discovery) and anything that contacts the app-server (gateway) will also be using ATTLS, they will see any HTTP traffic as HTTPS.

So, in this code, the metadata must claim that the connection is 100% HTTPS, 0% HTTP.
Because by all external observers that is what it is. It's just that ATTLS is the thing accomplishing that.

When writing that I checked and noticed I almost made a regression about that. So, fixed it https://github.com/zowe/zlux-server-framework/pull/580/commits/5f84184058540da40c7d6f6f1985e82f147dae7e

skurnevich (approved):
This works for me but only when there is no outbound AT-TLS rule for app-server set. 
I was able to make it full ATTLS for the app server by changing _makeMainInstanceProperties protocolObject to use isClientAttls
```
      httpEnabled: this.isClientAttls,
      httpsEnabled: !this.isClientAttls
```
In that case the app-server works with the outbound rule enabled, so it uses full ATTLS. However i was not able to properly test it with ZSS, so lets merge this one as there is a code freeze and continue in the different PR

PR #583, opened 2025-01-10
Use gateway client attls status in eureka object to determine http vs…

1000TurquoisePogs:
This PR builds upon https://github.com/zowe/zlux-server-framework/pull/580 by altering the eureka object's status of https/http true/false dependent upon the status of the gateway's client attls setting.

This is because, client attls alters whether you must send a request as 'http' or 'https' in order to avoid double-tls, or no-tls situations.
In this PR, the app-server reports itself as 'http' if the gateway has client tls enabled, so that when the gateway contacts the app-server, it will HTTPS, as desired.

This logic does not consider the state of the app-server's server ATTLS setting, because whether server ATTLS is enabled or disabled, the app-server should always be HTTPS.

### DIFF
--- a/lib/apiml.js
+++ b/lib/apiml.js
@@ -83,12 +83,12 @@ function ApimlConnector({ hostName, port, discoveryUrls,
   Object.assign(this, { hostName, port, discoveryUrls,
                         discoveryPort, tlsOptions, eurekaOverrides, isClientAttls });
   //TODO config should never be checked through env var, but is temporarily needed to temporarily read gateway's ATTLS state to provide it with Eureka info it can work with.
-  const clientGlobalAttls = process.env['ZWE_zowe_network_client_tls_attls'] == 'true';
-  const clientGatewayAttls = process.env['ZWE_components_gateway_zowe_network_client_tls_attls'] == 'true';
-  const clientAttls = clientGlobalAttls || clientGatewayAttls;
+  const clientGlobalAttls = process.env['ZWE_zowe_network_client_tls_attls'];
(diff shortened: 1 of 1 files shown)


lib/apiml.js
```diff
@@ -82,6 +82,19 @@ function ApimlConnector({ hostName, port, discoveryUrls,
                           discoveryPort, tlsOptions, eurekaOverrides, isClientAttls }) {
   Object.assign(this, { hostName, port, discoveryUrls,
                         discoveryPort, tlsOptions, eurekaOverrides, isClientAttls });
+  //TODO config should never be checked through env var, but is temporarily needed to temporarily read gateway's ATTLS state to provide it with Eureka info it can work with.
+  const clientGlobalAttls = process.env['ZWE_zowe_network_client_tls_attls'] == 'true';
+  const clientGatewayAttls = process.env['ZWE_components_gateway_zowe_network_client_tls_attls'] == 'true';
+  const clientAttls = clientGlobalAttls || clientGatewayAttls;
+  this.isGatewayClientAttls = false;
+  if ((clientGlobalAttls !== false) && (clientLocalAttls !== false) && (!clientAttls)) {
```
skurnevich: I can't see the `clientLocalAttls` definition is it a typo and should be `clientGatewayAttls`? 
In that case i do not quite understand this condition, i believe it will never pass as clientGlobalAttls and clientGatewayAttls both are defined booleans. if both are true then clientAttls will always be true.

1000TurquoisePogs: you're right, this was a mistake of me copying over from the util.js function.
it's not going to have the same logic.

skurnevich (approved):

PR #597, opened 2025-03-20
Added certificate checker utility

1000TurquoisePogs:
Depends upon https://github.com/zowe/zlux-app-server/pull/338

This PR has a tool, utils/certificateChecker.js
If you provide it with args of certificate files, file type, and alias, then it will try to load the cert and then check if it has expired and if it has the right EKU properties if appropriate.
It is intended to be used by the zowe startup process to prevent startup when known certificate incompatibilities are found.


timgerstel (approved):
LGTM

PR #600, opened 2025-03-28
Remove axios in favor of undici, a library made by nodejs

1000TurquoisePogs:
This PR switches out the `axios` use in apimlStorage.ts for `undici` instead. No functionality is changed, but this should fix a high memory utilization issue.

How to test:

Install sample-angular-app (zwe components install -o /zowe/components/app-server/share/sample-angular-app)
Enable it and the caching service
Ensure you dont have ZLUX_NO_CLUSTER=1  set. This storage doesnt work in no-cluster mode.
Open it up (need to use ?use-v2-desktop=true in URL)
On the storage panel, set the dropdown to "HA"
Set, get, and play around with the values to confirm its setting & getting data properly.


timgerstel (approved):
LGTM

PR #614, opened 2025-08-22
ipv6 for app-server

1000TurquoisePogs:
This PR acts upon strings printed to the users, eureka registration, proxying, and headers.

These strings take zowe configuration values for host/ip and create a URL from them in the form of `https://ip:port/`
This is incorrect when ipv6 ips are used, as they must be wrapped in `[`.
I detect this by detecting `:` in the ip.
Then I wrap it like `https://[ipv6]:port/`.

This same decision making is applied to such URLs in eureka registration (apiml.js)

Sometimes, the reverse is needed - zowe may pass in env vars with full URLs and when we just need the host, we must remove the '[]'.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to the Zlux Server Framework package will be documented in t
 This repo is part of the app-server Zowe Component, and the change logs here may appear on Zowe.org in that section.
 
 ## 3.4.0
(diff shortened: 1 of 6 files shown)


timgerstel (approved):
not sure why ipv6 needs to be wrapped in brackets but lgtm

1000TurquoisePogs:
ipv6 needs to be wrapped in brackets according to https://datatracker.ietf.org/doc/html/rfc2732#section-3

PR #616, opened 2025-09-08
MVD-6575: Add ability to deactivate Change Password in Desktop UI

shikharka:
- Added enablePasswordChange flag to zowe.yaml to control visibility of 'Change Password' option in the Personalization Panel.

- By default, the flag is set to true to preserve existing behavior. When explicitly set to false in zowe.yaml, the 'Change Password' option will be hidden from the Personalization Panel.


DivergentEuropeans (approved):

1000TurquoisePogs (approved):

1000TurquoisePogs:
@shikharka  you need to add this item to the schema
https://github.com/zowe/zlux-app-server/blob/v3.x/staging/schemas/app-server-config.json this should describe the property you are adding.

You also need to edit CHANGELOG.md with the description you'd like people to see on our release notes about what property was added and what it does.

PR #635, opened 2025-11-19
Upgraded js-yaml to 4.1.1

ChongZhou-Broadcom:
## Proposed changes

This PR addresses Issue: [*Link to Github issue within https://github.com/zowe/zlux/issues* if any]

N/A

This PR depends upon the following PRs:

N/A

## Type of change
Please delete options that are not relevant.
- [x] Chore, repository cleanup, updates the dependencies.

## PR Checklist
Please delete options that are not relevant.
- [x] If the changes in this PR are meant for the next release / mainline, this PR targets the "staging" branch.
- [x] My code follows the style guidelines of this project (see: [Contributing guideline](https://github.com/zowe/zlux/blob/master/CONTRIBUTING.md))
- [x] Relevant update to CHANGELOG.md

## Testing

Manually installed the component pax into the installation in my test lpar and did smoke tests. No burnt smell came out.

## Further comments

- Upgraded `js-yaml` in the root/package-lock.json from `4.1.0 to 4.1.1` as suggested by Blackduck.

PR #637, opened 2025-11-27
Upgraded js-yaml to 4.1.1

ChongZhou-Broadcom:
Port the change (#635 ) from v3/staging to v2/staging.


ChongZhou-Broadcom:
I am closing this PR because the dependency `js-yaml` is already seen to be 4.1.1.

I pulled the latest states of v2.x-staging into my local repo and ran `npm list js-yaml`, it showed the version of `js-yaml` is 4.1.1.

PR #646, opened 2025-12-21
TS for util.js

1000TurquoisePogs:
This PR is a quick convert to typescript, to allow future typescript enhancements.
It is intended to be used alongside index.js - it will not work with index.ts because of the export statement differences.

### DIFF
--- a/lib/util.ts
+++ b/lib/util.ts
@@ -340,7 +340,7 @@ const errorProto = {
     "messageDetails": "An error occurred when processing the request"
   };
 
-export function  makeError(details) {
+export function makeErrorObject(details) {
   if ((details._objectType !== undefined) 
       || (details._metaDataVersion !== undefined)) {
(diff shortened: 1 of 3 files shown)


lib/util.ts
```diff
@@ -17,17 +17,18 @@ if (!global.COM_RS_COMMON_LOGGER) {
...
-const Promise = require('bluebird');
-const ipaddr = require('ipaddr.js');
-const dns = require('dns');
+import * as path from 'node:path';
+import * as fs from 'node:fs';
+import * as util from 'node:util';
+import * as BBPromise from 'bluebird';
+import * as ipaddr from 'ipaddr.js';
+import * as dns from 'node:dns';
+import * as mergeUtils from '../utils/mergeUtils';
```
Martin-Zeithaml, line 26: `mergeUtils` not used.

PR #681, opened 2026-05-13
Remove accept-language-parser

1000TurquoisePogs:
## Proposed changes

Replace the `accept-language-parser` npm dependency with an equivalent inline implementation in `lib/translation-utils.js`. The replacement adds security hardening that was absent in the library: input length capping, fail-closed validation on malformed entries, and strict BCP 47 subtag character and length checks. A formal mocha/chai test suite covering both functional behavior and adversarial inputs is added alongside the change.

## Testing

Run the new unit tests:

```
cd zlux-server-framework
npm run test:translation
```

Manual verification: start the app-server and change the browser language preference. The Zowe Desktop plugin labels should translate correctly according to available `pluginDefinition.i18n.*.json` files.

### Why the library is no longer needed

`accept-language-parser` provides two exports: `parse()` and `pick()`. Only `pick()` is used, in a single call site in `lib/translation-utils.js`. The call takes:

- `availableTranslations` — a plain string array of language tags derived from filenames on disk (e.g. `['en', 'es-ES', 'fr']`)
- `acceptLanguage` — the raw `Accept-Language` HTTP header string

The algorithm `pick()` implements is straightforward: split the header on commas, extract each language tag and its optional `q=` quality value, sort by quality descending, and return the first supported tag whose language code and optional region match a preference entry.

The inline `pickLanguage()` function replicates this logic exactly and handles the same cases the library does within this codebase's usage pattern.

### Where the replacement is better for our needs

The library performs no validation on the incoming header string. A malicious client could send an arbitrarily large or structurally bizarre `Accept-Language` header. The inline implementation adds the following defenses that were absent in the library:

| Check | Behaviour on failure |
|---|---|
| Header length > 256 characters | Return `null` immediately — no parsing |
| Empty entry after splitting on `,` (trailing/double commas) | Return `null` — treat whole header as malformed |
| More than one `;`-delimited segment per entry | Return `null` |
| Subtag fails `[a-zA-Z0-9]` character check | Return `null` |
| Subtag length > 8 characters (BCP 47 maximum) | Return `null` |
| More than 3 subtags in a language tag | Return `null` |
| Quality value not prefixed with `q=` | Return `null` |
| Quality value outside `[0, 1]` or non-numeric | Return `null` |

All failures are fail-closed: `pickLanguage()` returns `null`, which causes `translate()` to return the untranslated plugin definition — the same safe default that applied when the library found no matching language.

### What the library handled that is intentionally excluded

The library supported 3-part language tags with a script subtag (`zh-Hant-TW`). The subtag validation in `pickLanguage()` permits up to 3 subtags, so such tags parse correctly. However, the server's own translation files only use `code` or `code-region` format (derived from filenames like `pluginDefinition.i18n.es-ES.json`), so script-subtag matching is not exercised in practice.

The library also accepted `*` as a wildcard language tag. This is not used by any browser in a way that would affect translation selection against a small fixed list of supported languages, and accepting `*` as valid input from untrusted clients is undesirable, so it is intentionally not supported.


RamkrishnaRocket1 (approved):
Tested and the languages component is working fine Tested on Chrome Firefox and Edge

PR #683, opened 2026-05-13
Replace unidici with builtin https for memory fix

1000TurquoisePogs:
## Proposed changes

Replace the `undici` npm dependency in `apimlStorage.ts` with Node.js's built-in `https` module, eliminating ~3.7 GB of committed memory on z/OS at startup. Add a self-contained mock APIML Caching Service and restructure the storage tests so they run against the mock when no real APIML gateway is available.

### `lib/apimlStorage.ts` — remove undici, use native `https`

**Problem:** Importing `undici` at module load time caused ~3.7 GB of memory to be requested immediately on z/OS. The same issue previously existed with `axios` (removed in v3.2.0) and has now recurred with `undici`.

**Fix:** Replace `undici.Client` with `https.Agent` from Node.js's built-in `https` module.

- `configure()` now creates an `https.Agent` with `keepAlive: true`, passing the caller's TLS options directly. All fields used by zlux (`cert`, `key`, `ca`, `crl`, `ciphers`, `secureOptions`, `secureProtocol`, `ecdhCurve`, `rejectUnauthorized`, `checkServerIdentity`, `enableTrace`) are natively supported by `https.Agent` with no conversion.
- `apimlDoRequest()` uses `https.request()`. Request bodies are serialized to a `Buffer` once via `Buffer.from(JSON.stringify(...))` — this encodes UTF-8 exactly once, avoids the double-encode that would occur with a string, and `Buffer.length` is a zero-cost property read for `Content-Length`.
- Response bodies are collected via a `collectResponseBody()` helper that reads chunks into a `Buffer[]`, concatenates, and decodes once.
- Connection pooling is preserved through `keepAlive: true` on the agent.
- The native `https` module is already loaded by the rest of the server, so this import adds zero additional memory overhead.

**Consequences:**
- Memory at startup on z/OS is expected to drop from ~4 GB to under 1GB.
- No change in observable behavior: all HTTP methods, TLS options, error codes, and response parsing are identical.
- `undici` is removed from `package.json` dependencies.

### `test/storage/` — mock server and restructured tests

**Problem:** The existing tests in `apimlStorage.js` and `apimlStorageViaUnified.js` skipped entirely when `GATEWAY_HOST`/`GATEWAY_PORT`/`CLIENT_KEY`/`CLIENT_CER` env vars were absent, meaning CI and local development got zero test coverage for this module.

**Fix:** Add a mock APIML Caching Service and restructure the tests to fall back to it automatically.

[...]

PR #689, opened 2026-05-15
Chore/v2/use builtin request

1000TurquoisePogs:
This PR is the same as https://github.com/zowe/zlux-server-framework/pull/683 but for v2.
As its using builtin node features, it's going to be acceptable for v2 same as in v3.


RamkrishnaRocket1 (approved):
Tested no dependency of axios library seen

PR #690, opened 2026-05-15
Set specific versions and enforce it in workflow

1000TurquoisePogs:
This PR sets all package.json versions to what was found in package-lock.json to set a baseline for specific versions rather than relying on version ranges.

Next, I've added the workflow action zowe-actions/shared-actions/validate-package-json@main
enforces this and also enforces that packages are at least a week old.

### DIFF
--- a/.github/workflows/build-core.yml
+++ b/.github/workflows/build-core.yml
@@ -136,7 +136,13 @@ jobs:
     needs: check-permission
     steps:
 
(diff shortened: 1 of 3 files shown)


timgerstel (approved):

1000TurquoisePogs:
The build passed but github itself is broken, so I must force-merge this https://github.com/zowe/zlux-server-framework/actions/runs/25923189121/job/76197382429

PR #696, opened 2026-06-09
fix: guard against null agent request options in proxy setup

struga0258:
Problem
[zluxUtil.getAgentRequestOptions()](vscode-file://vscode-app/c:/Users/jstruga/AppData/Local/Programs/Microsoft%20VS%20Code/6a44c352bd/resources/app/out/vs/code/electron-browser/workbench/workbench.html) can return null when the agent host/port cannot be resolved for a given URL prefix (e.g., misconfigured agent settings or missing ZSS configuration). When this happens, the code immediately tries to access [requestOptions.host](vscode-file://vscode-app/c:/Users/jstruga/AppData/Local/Programs/Microsoft%20VS%20Code/6a44c352bd/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [requestOptions.port](vscode-file://vscode-app/c:/Users/jstruga/AppData/Local/Programs/Microsoft%20VS%20Code/6a44c352bd/resources/app/out/vs/code/electron-browser/workbench/workbench.html), resulting in an unhandled TypeError crash that brings down the app server.

Solution
Added a null-check on the return value of [getAgentRequestOptions()](vscode-file://vscode-app/c:/Users/jstruga/AppData/Local/Programs/Microsoft%20VS%20Code/6a44c352bd/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
If null, log a warning (ZWED0005W) identifying the problematic URL prefix and skip proxy setup for that route
Populated the previously-empty ZWED0005W message placeholder with a descriptive warning string


1000TurquoisePogs:
So if you have invalid config today the server crashes.
This changes it so that the server doesn't crash.
But, should the server really continue to run if the config is invalid?

PR #704, opened 2026-08-03
Verify RBAC is enabled for instance and site scope plugin configuration updating

skurnevich:
## Proposed changes

Verify RBAC is enabled for instance and site scope plugin configuration writing and deleting.


1000TurquoisePogs (dismissed):
Thank you

1000TurquoisePogs (approved):

PR #705, opened 2026-08-03
Remove dead encryption lib

1000TurquoisePogs:
the 'encryption.js' file was a trivial wrapper for `node:crypto` functions, some of which were removed in node v22, after years of being deprecated. It's long past time to move on from this file, which is unused in zowe itself.


struga0258 (approved):

1000TurquoisePogs:
This PR is needed for another PR to succeed, and that PR is preventing build, so one needs to be force merged to resolve.

PR #709, opened 2026-08-06
Fix WebSocket RBAC bypass in ZSS auth handler

timgerstel:
## Proposed changes

This PR fixes a WebSocket authorization bypass in the ZSS (`sso-auth`) handler.

`webauth.middleware` sets `syncOnly = req.url.endsWith(".websocket")`. When ZSS is the active auth handler and RBAC is enabled, `ZssHandler.authorized()` reached the `syncOnly` branch and unconditionally set `result.authorized = true` for any authenticated user, skipping the `/saf-auth` SAF check entirely. As a result, every WebSocket-based dataservice (terminal-proxy, ws proxies, plugin ws routers) was reachable by any authenticated user regardless of their SAF profile permissions.

Because `authorized()` is an async coroutine that is `yield`ed (awaited) by the auth middleware, and `_callAgent()` returns a promise, the real SAF authorization check can be performed for WebSocket requests too. This PR removes the fail-open `syncOnly` branch so the code proceeds to the actual `/saf-auth` check, and removes the now-unused `syncOnly` destructuring. Any error in the surrounding `try/catch` now fails closed (`authorized = false`).

CVSS 3.1 - 8.1 :: AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:H/A:N

This PR addresses Issue: [*Link to Github issue within https://github.com/zowe/zlux/issues* if any]

This PR depends upon the following PRs: N/A

## Type of change

## PR Checklist
- [x] I have commented my code, particularly in hard-to-understand areas

## Testing

With `components.app-server.dataserviceAuthentication.rbac: true` and the ZSS/`sso-auth` handler active:

1. Log in as a user who does NOT have the SAF profile READ access for a WebSocket-backed dataservice (e.g. terminal-proxy).
2. Attempt to open the WebSocket connection (URL ending in `.websocket`).
   - Before this fix: the connection is authorized (bypass) despite lacking SAF access.
   - After this fix: the `/saf-auth` check runs; the request is rejected (403) when the user lacks the profile.
3. Log in as a user who DOES have READ access to the profile and confirm the WebSocket connection is authorized.
4. With `rbac: false` (default), confirm WebSocket dataservices still work unchanged (authorization is short-circuited earlier via `bypassAuthorizatonCheck`).

## Further comments

The `syncOnly` flag was originally used to skip the RBAC check on the assumption that an async agent call could not be made during WebSocket handling. That assumption is not accurate here: the auth middleware already awaits `authorized()`, so the SAF call completes correctly for WebSocket requests. The fix keeps behavior identical when RBAC is disabled and only changes the RBAC-enabled WebSocket path from failing open to enforcing the SAF check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to the Zlux Server Framework package will be documented in t
 This repo is part of the app-server Zowe Component, and the change logs here may appear on Zowe.org in that section.
 
 ## 3.6.0
+- Security: Fixed a WebSocket authorization bypass in the ZSS auth handler. When RBAC was enabled, WebSocket dataservice requests skipped the SAF authorization check and were granted to any authenticated user regardless of their SAF profile permissions. The SAF check is now enforced for WebSocket requests.
 - Enhancement: The `/server/environment` response now includes `agent.host`, exposing the effective ZSS/agent hostname (the same value previously returned by the `/server/proxies` endpoint, including the APIML gateway hostname override when agent mediation layer is enabled). This allows plugins to retrieve the agent host through the authenticated `ZoweZLUX.environment.getAgentHost()` API.
 - Bugfix: Removed the unauthenticated `/server/proxies` endpoint. Its only consumers (the TN3270 and VT terminal apps) have been updated to use `ZoweZLUX.environment.getAgentHost()` instead.
 - Moved `trivial-auth` and `internal-auth` authentication plugins to `devPlugins/` directory. These plugins are development/troubleshooting tools that should not ship in production distributions. [(#700)](https://github.com/zowe/zlux-server-framework/pull/700) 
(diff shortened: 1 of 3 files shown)


timgerstel:
Fixes: https://github.com/zowe/security-reports/issues/2540

timgerstel:
Thoughts on why this fix is valid:

The middleware ordering guarantees auth runs before the plugin's ws handler. In webapp.js:1694 _makeRouter builds: 

`[...pluginChain, injectServiceDef, auth.middleware, ...otherMw, router]`

and the plugin's router.ws('/') handler (which calls handler.createSession(req) and wires ws.on('message')) is pushed last.

1000TurquoisePogs:
Closing this as redundant to https://github.com/zowe/zlux-server-framework/pull/710

PR #710, opened 2026-08-11
Enforce RBAC for WebSocket dataservices

struga0258:
## Summary
Fixes two RBAC gaps found in pen testing. WebSocket dataservices no longer bypass the SAF check, the ZSS handler's `syncOnly` short circuit is removed, so WebSocket requests are authorized the same way as REST requests and the socket is closed if the user isn't permitted. Separately, the config dataservice now rejects PUT/DELETE to `instance`/`site` scopes with 403 (ZWED0145E) unless RBAC is enabled, preventing low-privilege users from overwriting server-wide or site-wide plugin configuration.


1000TurquoisePogs (dismissed):
This has been demonstrated to me - RBAC permit/deny of GET requests to the terminal proxy now will close the terminal connection.

1000TurquoisePogs (approved):

